PR #3, opened 2018-12-08
Add ability to configure where draco is installed

mkrus:
Support setting DRACOSDK env var to point to where Draco include and
lib folders are

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ for it, run qmake again like this:
 
     qmake kuesa.pro -- --draco=no
 
+If Draco is not installed in the default location on your system, you can
+set the DRACOSDK environment variable to point to where Draco is installed.
+Use the DRACOSDK_LIBS environment variable to point to the folder containing
+the Draco libraries.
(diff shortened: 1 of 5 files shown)


kuesa.pro
```diff
@@ -24,14 +24,13 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-lessThan(QT_MAJOR_VERSION, 5) | lessThan(QT_MINOR_VERSION, 11): error("This project requires Qt 5.11")
+lessThan(QT_MAJOR_VERSION, 5) | lessThan(QT_MINOR_VERSION, 12): error("This project requires Qt 5.12")
```
jjcasmar: Should this be a different patch?

krf: https://github.com/KDAB/kuesa/pull/6

mkrus: yeah, removed the change here

PR #5, opened 2018-12-10
Compress draco

jjcasmar:
Can you please have a look? I have checked all the data structures im generating, but still cant encode it.

### DIFF
--- a/tools/assetpipelineeditor/dracocompressor.cpp
+++ b/tools/assetpipelineeditor/dracocompressor.cpp
@@ -1,0 +1,166 @@
+#include "dracocompressor.h"
+
+#include <Qt3DRender/QGeometry>
+#include <Qt3DRender/QAttribute>
+#include <Qt3DRender/QBuffer>
+#include <draco/mesh/mesh.h>
+#include <draco/attributes/geometry_attribute.h>
+#include <draco/attributes/geometry_indices.h>
+#include <QVector>
+
+namespace
+{
+draco::GeometryAttribute::Type attributeTypeFromName(const QString &attributeName)
+{
+    auto type = draco::GeometryAttribute::Type::INVALID;
+
+    if (attributeName == Qt3DRender::QAttribute::defaultPositionAttributeName())
+        type = draco::GeometryAttribute::Type::POSITION;
+    if (attributeName == Qt3DRender::QAttribute::defaultNormalAttributeName())
+        type = draco::GeometryAttribute::Type::NORMAL;
+    if (attributeName == Qt3DRender::QAttribute::defaultTextureCoordinate1AttributeName())
+        type =  draco::GeometryAttribute::Type::TEX_COORD;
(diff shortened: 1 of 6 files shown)


tools/assetpipelineeditor/dracocompressor.cpp
```diff
@@ -0,0 +1,175 @@
...
+                    << data[i];
+        qDebug() << "---End----";
+
+        draco::GeometryAttribute meshAttribute;
+        meshAttribute.Init(dracoAttributeType,
+                           dracoBuffer,
+                           attribute->vertexSize(),
+                           dracoDataType,
+                           false,
+                           attribute->byteStride(),
```
mkrus: stride and offset should both be 0 since your copying data into separate new buffers

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -337,6 +344,43 @@ void MainWindow::openFile()
...
+        const QFileInfo gltfFileInfo(m_filePathURL);
+        const QDir gltfFileDir = gltfFileInfo.dir();
+        QFile compressedBufferFile(
+                QUrl(gltfFileDir.filePath({ "compressedBuffer.bin" })).toLocalFile());
+        if (compressedBufferFile.open(QIODevice::WriteOnly))
+            compressedBufferFile.write(compressedBufferData);
+
+        QJsonObject compressedBufferObject;
+        compressedBufferObject["byteLength"] = compressedBufferData.size();
+        compressedBufferObject["uri"] = QUrl(gltfFileDir.filePath({ "compressedBuffer.bin" })).toLocalFile();
```
mkrus: should be relative path

tools/assetpipelineeditor/dracocompressor.cpp
jjcasmar: I cant see why. Im creating a copy of the QBuffer into the dracoBuffer, so I think stride and offset should be the same, no?

tools/assetpipelineeditor/mainwindow.cpp
jjcasmar: Sure, just doing some tests. Now that Ive finally fix the attribute generation, I will address this things

jcelerier:
hi all, two questions about this PR : 

- shouldn't the `bufferView` be updated in the saved json to target the new buffer ? 
- from what I understood from the format, it is possible to have a `buffer` that looks like : 

     [ - - - whatever - - - ] [ an uncompressed mesh ] [ - - - whatever - - - ]

If the point of implementing draco is to reduce the overall asset size, shouldn't a new binary file be copied from the existing one where the mesh resides, but with the uncompressed slice becoming compressed ? (of course this requires adjusting all the bufferView indices relating to this buffer after the compressed mesh). Because else this means that the original .bin file will still have to be carried around, if for instance it packs both the mesh and the textures of a node.

PR #10, opened 2018-12-19
Add Scene3D based Kuesa application template for QtCreator

lemirep:
Task-Id: KUE-230


tools/qtc_templates/kuesa/project/empty_quick/app.pro
```diff
@@ -0,0 +1,19 @@
+QT += quick
+CONFIG += resources_big
+
+DEFINES += QT_DEPRECATED_WARNINGS
+
+SOURCES += \\
```
er-vin: Shouldn't that be just one \ ?

In fact, I'm not quite sure how qmake would deal with a double one here. :-)

tools/qtc_templates/kuesa/project/empty_quick/app.pro
```diff
@@ -0,0 +1,19 @@
...
+        %{CppFileName}
```
er-vin: Indent with only four spaces here? That's what you used for the main.qml below.

tools/qtc_templates/kuesa/project/empty_quick/main.qml
```diff
@@ -0,0 +1,67 @@
...
+
+Item {
+    id: root
+
+    Scene3D {
+        id: scene3d
+        anchors.fill: parent
+        aspects: ["render", "input", "logic", "animation"]
+
+        Kuesa.SceneEntity {
```
er-vin: I'm still pondering about that one... wouldn't it be a better practice to put this entity in its own qml file? This way we'd split between QtQuick and Kuesa code. It's what we seem to do most of the time AFAIK.

tools/qtc_templates/kuesa/project/empty_quick/main.qml
```diff
@@ -0,0 +1,67 @@
...
+                InputSettings { },
+                DirectionalLight {
+                    worldDirection: frameGraph.camera.viewVector
+                    intensity: 0.8
+                    color: "white"
+                }
+            ]
+
+            Camera {
+                id: mainCamera
```
er-vin: I'd rename that to "fallbackCamera" since it's mostly just in case the export wouldn't provide something, isn't it?

lemirep: depends if you are using the GLTFLoader (by specifying a path in the wizard) or not I suppose, this would also have to be addressed in the other template.

tools/qtc_templates/kuesa/project/empty_quick/app.pro
lemirep: done

tools/qtc_templates/kuesa/project/empty_quick/app.pro
lemirep: done

tools/qtc_templates/kuesa/project/empty_quick/main.qml
lemirep: done

tools/qtc_templates/kuesa/project/empty_quick/SceneEntity.qml
```diff
@@ -0,0 +1,55 @@
...
+import Kuesa %{KuesaVersion} as Kuesa
+
+Kuesa.SceneEntity {
+    id: root3D
+
+    components: [
+        RenderSettings {
+            activeFrameGraph: Kuesa.ForwardRenderer {
+                id: frameGraph
+                camera: root3D.cameras.names.length ? root3D.cameras.find(root3D.cameras.names[0]) : fallBackCamera
```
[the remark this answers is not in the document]
lemirep: done and updated other template as well

PR #14, opened 2019-01-10
Cleanup effects

mkrus:
Hack in no longer needed with change in qt5.12. Actually wonder if it should be conditionally compiled in if qt ver is < 5.12.2?

Also includes some clang-tidy formatting changes.

### DIFF
--- a/src/core/metallicroughnesseffect.cpp
+++ b/src/core/metallicroughnesseffect.cpp
@@ -357,8 +357,10 @@ MetallicRoughnessEffect::MetallicRoughnessEffect(Qt3DCore::QNode *parent)
     // don't try to sample from them.
     // Can probably go away once we generate the shaders and deal in this
     // case in a better way.
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
     addParameter(new QParameter(QStringLiteral("envLight.irradiance"), new QTexture2D));
     addParameter(new QParameter(QStringLiteral("envLight.specular"), new QTexture2D));
+#endif
 
     QMetaObject::invokeMethod(this, "initVertexShader", Qt::QueuedConnection);
     m_invokeInitVertexShaderRequested = true;
--- a/tests/manual/forwardrenderer_scene/main.qml
+++ b/tests/manual/forwardrenderer_scene/main.qml
@@ -51,8 +51,7 @@ Entity {
         }
     ]
 
-    function enablePostProcessingEffect(effect, enable)
-    {
+    function enablePostProcessingEffect(effect, enable) {
         if (enable)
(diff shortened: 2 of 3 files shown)


src/core/metallicroughnesseffect.cpp
```diff
@@ -205,14 +205,14 @@ MetallicRoughnessEffect::MetallicRoughnessEffect(Qt3DCore::QNode *parent)
...
 {
-    const auto enabledLayers = QStringList{ QStringLiteral("noBaseColorMap"),
-                                            QStringLiteral("noMetalRoughMap"),
-                                            QStringLiteral("noAmbientOcclusionMap"),
-                                            QStringLiteral("noEmissiveMap"),
-                                            QStringLiteral("noNormalMap"),
-                                            QStringLiteral("noHasColorAttr"),
-                                            QStringLiteral("noDoubleSided"),
-                                            QStringLiteral("noHasAlphaCutoff") };
+    const auto enabledLayers = QStringList { QStringLiteral("noBaseColorMap"),
```
er-vin: Hm... this was done this way by clang-format. Are you sure about that format change?

src/core/metallicroughnesseffect.cpp
```diff
@@ -351,15 +351,6 @@ MetallicRoughnessEffect::MetallicRoughnessEffect(Qt3DCore::QNode *parent)
...
 
-    // Note that even though those parameters are not exposed in the API,
-    // they need to be kept around for now due to a bug in some drivers/GPUs
-    // (at least Intel) which cause issues with unbound textures even if you
-    // don't try to sample from them.
-    // Can probably go away once we generate the shaders and deal in this
-    // case in a better way.
-    addParameter(new QParameter(QStringLiteral("envLight.irradiance"), new QTexture2D));
-    addParameter(new QParameter(QStringLiteral("envLight.specular"), new QTexture2D));
-
```
er-vin: I'm not sure we can remove it now... I mean due to ABI compatibility it could run on any 5.12.x version and only 5.12.2 will carry the fix. I'd be tempted to make that conditional based on 5.13 being detected at build time.

mkrus: done

tests/manual/forwardrenderer_scene/main.qml
```diff
@@ -51,8 +51,7 @@ Entity {
         }
     ]
 
-    function enablePostProcessingEffect(effect, enable)
-    {
+    function enablePostProcessingEffect(effect, enable) {
```
jjcasmar: Unrelated?

jjcasmar: Dont think so

PR #15, opened 2019-01-10
Fix multiline string not accepted by gcc

chsterz:
This fixes an error about `#version` not being a proper preprocessor directive


chsterz:
Just realized it is not happening with my home-desktop (with even the same g++ version)
thinking of pulling that back.

jcelerier:
the original one worked fine for me with g++ 8.2.1 too

PR #16, opened 2019-01-11
Compress draco 2

jcelerier:
This pull request starts to have something that looks like a working implementation (see https://github.com/KDAB/kuesa/pull/16/files?w=1 to get a diff without the whitespace changes - github really ought to have a nice button for this).

### What works
* Compressing and reloading these in Kuesa: 
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Box/glTF
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Cube/glTF
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedCube/glTF
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxAnimated/glTF
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Duck/glTF
  - https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Buggy/glTF
  - The DodgeViper example

### What doesn't
* Original buffers aren't stripped of uncompressed data properly yet (it works if there is no interleaving of multiple bufferViews) - edit : actually it looks like this works (at least the interleaved glTF example works : https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxInterleaved ) ; I suppose that this is because once we're in the Qt3D world (as the mesh data are), there is no interleaving anymore ? 

* Currently, all the indices relative to the following bufferViews in the whole GLTF file (in known extensions) are readjusted. But if someone has a custom extension that also uses bufferViews to access data, we won't have any way to know that we have to adjust them and this will break the custom extension. (and we can't assume that it will be exposed as `"bufferView": 123` to try and replace them, it may be a different key like the ones used in `"primitives"`).

### What's missing
* tests, prettyfication, proper comments and documentation
* support for embedded data (depends on https://github.com/KDAB/kuesa/pull/25)

### DIFF
--- a/src/core/gltf2exporter/dracocompressor_p.cpp
+++ b/src/core/gltf2exporter/dracocompressor_p.cpp
@@ -1,0 +1,373 @@
+/*
+    dracocompressor_p.cpp
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Juan Jose Casafranca <juan.casafranca@kdab.com>
+
+    Licensees holding valid proprietary KDAB Kuesa licenses may use this file in
+    accordance with the Kuesa Enterprise License Agreement provided with the Software in the
+    LICENSE.KUESA.ENTERPRISE file.
+
+    Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as
+    published by the Free Software Foundation, either version 3 of the
+    License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <draco/mesh/mesh.h>
+#include <draco/attributes/geometry_attribute.h>
+#include <draco/attributes/geometry_indices.h>
+#include <draco/io/mesh_io.h>
+#include "dracocompressor_p.h"
+#include "gltf2exporter_p.h"
+#include <Qt3DRender/QGeometry>
+#include <Qt3DRender/QAttribute>
+#include <Qt3DRender/QBuffer>
+#include <QVector>
+#include <QFile>
+#include <QLoggingCategory>
+#include <QDataStream>
+#include <fstream>
+#include <iterator>
+#include <iostream>
+
+QT_BEGIN_NAMESPACE
+Q_LOGGING_CATEGORY(dracocompressor, "DracoCompressor")
+namespace Kuesa {
+namespace DracoCompression {
+namespace {
+draco::GeometryAttribute::Type attributeTypeFromName(const QString &attributeName)
+{
+    auto type = draco::GeometryAttribute::Type::INVALID;
+
+    if (attributeName == Qt3DRender::QAttribute::defaultPositionAttributeName())
+        type = draco::GeometryAttribute::Type::POSITION;
+    else if (attributeName == Qt3DRender::QAttribute::defaultNormalAttributeName())
+        type = draco::GeometryAttribute::Type::NORMAL;
+    else if (attributeName == Qt3DRender::QAttribute::defaultColorAttributeName())
+        type = draco::GeometryAttribute::Type::COLOR;
+    else if (attributeName == Qt3DRender::QAttribute::defaultTextureCoordinateAttributeName())
+        type = draco::GeometryAttribute::Type::TEX_COORD;
+    else if (attributeName == Qt3DRender::QAttribute::defaultTextureCoordinate1AttributeName())
+        type = draco::GeometryAttribute::Type::TEX_COORD;
+    else if (attributeName == Qt3DRender::QAttribute::defaultTextureCoordinate2AttributeName())
+        type = draco::GeometryAttribute::Type::TEX_COORD;
+    else if (attributeName == Qt3DRender::QAttribute::defaultTangentAttributeName())
+        type = draco::GeometryAttribute::Type::GENERIC;
+    else if (attributeName == Qt3DRender::QAttribute::defaultJointWeightsAttributeName())
+        type = draco::GeometryAttribute::Type::GENERIC;
+    else if (attributeName == Qt3DRender::QAttribute::defaultJointIndicesAttributeName())
+        type = draco::GeometryAttribute::Type::GENERIC;
+
+    return type;
+}
+
+template<typename T>
+struct attribute_type_trait {
+    static constexpr draco::DataType type = draco::DataType::DT_INVALID;
+};
+
+template<>
+struct attribute_type_trait<float> {
+    static constexpr draco::DataType type = draco::DataType::DT_FLOAT32;
+};
+
+template<>
+struct attribute_type_trait<char> {
+    static constexpr draco::DataType type = draco::DataType::DT_INT8;
+};
+
+template<>
+struct attribute_type_trait<unsigned char> {
+    static constexpr draco::DataType type = draco::DataType::DT_UINT8;
+};
+
+template<>
+struct attribute_type_trait<unsigned short> {
+    static constexpr draco::DataType type = draco::DataType::DT_UINT16;
+};
+
+template<>
+struct attribute_type_trait<short> {
+    static constexpr draco::DataType type = draco::DataType::DT_INT16;
+};
+
+template<>
+struct attribute_type_trait<int> {
+    static constexpr draco::DataType type = draco::DataType::DT_INT32;
+};
+
+template<>
+struct attribute_type_trait<unsigned int> {
+    static constexpr draco::DataType type = draco::DataType::DT_UINT32;
+};
+
+template<>
+struct attribute_type_trait<double> {
+    static constexpr draco::DataType type = draco::DataType::DT_FLOAT64;
+};
+
+template<typename T, typename Sz>
+static constexpr Sz actualStride(Sz glStride, Sz vertexSize)
+{
+    return glStride == 0 ? vertexSize * Sz(sizeof(T)) : glStride;
+}
+
+template<typename T>
+QByteArray packedDataInBuffer(const QByteArray &inputBuffer,
+                              std::size_t vertexSize,
+                              std::size_t offset,
+                              std::size_t stride,
+                              std::size_t count)
+{
+    const auto elementSize = vertexSize * sizeof(T);
+
+    Q_ASSERT_X(count < std::numeric_limits<int>::max() / elementSize, "Kuesa::packedDataInBuffer", "Too many elements for a QByteArray");
+
+    QByteArray outputBuffer(int(count * elementSize), Qt::Uninitialized);
+
+    auto byteMarkerPosition = offset;
+    stride = actualStride<T>(stride, vertexSize);
+
+    for (std::size_t i = 0; i < count; ++i) {
+        std::memcpy(&(outputBuffer.data()[i * elementSize]), &(inputBuffer.data()[byteMarkerPosition]), elementSize);
+        byteMarkerPosition += stride;
+    }
+
+    return outputBuffer;
+}
+
+template<typename T>
+std::pair<QString, int> addAttributeToMesh(const Qt3DRender::QAttribute &attribute, draco::Mesh &mesh)
+{
+    const auto dracoAttributeType = attributeTypeFromName(attribute.name());
+    const auto dracoDataType = attribute_type_trait<T>::type;
+
+    if (dracoAttributeType == draco::GeometryAttribute::Type::POSITION)
+        mesh.set_num_points(attribute.count());
+
+    const auto attributeBuffer = attribute.buffer();
+
+    const QByteArray packedData = packedDataInBuffer<T>(attributeBuffer->data(),
+                                                        attribute.vertexSize(),
+                                                        attribute.byteOffset(),
+                                                        attribute.byteStride(),
+                                                        attribute.count());
+
+    draco::PointAttribute meshAttribute;
+    meshAttribute.Init(attributeTypeFromName(attribute.name()),
+                       nullptr,
+                       static_cast<int8_t>(attribute.vertexSize()),
+                       dracoDataType,
+                       false,
+                       actualStride<T>(attribute.byteStride(), attribute.vertexSize()),
+                       0);
+
+    auto attId = mesh.AddAttribute(meshAttribute, true, attribute.count());
+    if (attId == -1)
+        return { {}, -1 };
+
+    mesh.attribute(attId)->buffer()->Write(0, packedData.data(), static_cast<std::size_t>(packedData.size()));
+
+    return { attribute.property("semanticName").toString(), attId };
+}
+
+template<typename T>
+QVector<draco::PointIndex> createIndicesVectorFromAttribute(const Qt3DRender::QAttribute &attribute)
+{
+    const auto byteOffset = attribute.byteOffset();
+    const auto data = attribute.buffer()->data().mid(int(byteOffset));
+    const auto attrCount = attribute.count();
+    const int stride = actualStride<T>(int(attribute.byteStride()), 1);
+
+    Q_ASSERT_X(attrCount < std::numeric_limits<int>::max(), "Kuesa::createIndicesVectorFromAttribute", "Too many elements for a QVector");
+
+    QVector<draco::PointIndex> indicesVector(static_cast<int>(attrCount));
+
+    for (int i = 0, n = int(attrCount); i < n; ++i)
+        indicesVector[i] = *reinterpret_cast<const T *>(data.data() + i * stride);
+
+    return indicesVector;
+}
+
+void addFacesToMesh(const QVector<draco::PointIndex> &indices, draco::Mesh &dracoMesh)
+{
+    for (int i = 0, nbIndices = indices.size() / 3; i < nbIndices; ++i)
+        dracoMesh.AddFace({ indices[3 * i + 0],
+                            indices[3 * i + 1],
+                            indices[3 * i + 2] });
+}
+
+template<typename T>
+bool compressAttribute(
+        const Qt3DRender::QAttribute &attribute,
+        draco::Mesh &dracoMesh,
+        std::vector<std::pair<QString, int>>& attributes)
+{
+    auto compressedAttr = addAttributeToMesh<T>(attribute, dracoMesh);
+    if (compressedAttr.second == -1) {
+        qCWarning(dracocompressor) << "Could not add attribute to mesh: " << attribute.name();
+        return false;
+    }
+    attributes.push_back(compressedAttr);
+    return true;
+}
+
+Q_DECL_RELAXED_CONSTEXPR
+draco::GeometryAttribute::Type dracoAttribute(GLTF2ExportConfiguration::MeshAttribute m)
+{
+    switch (m) {
+    case GLTF2ExportConfiguration::Position:
+        return draco::GeometryAttribute::POSITION;
+    case GLTF2ExportConfiguration::Normal:
+        return draco::GeometryAttribute::NORMAL;
+    case GLTF2ExportConfiguration::TextureCoordinate:
+        return draco::GeometryAttribute::TEX_COORD;
+    case GLTF2ExportConfiguration::Color:
+        return draco::GeometryAttribute::COLOR;
+    case GLTF2ExportConfiguration::Generic:
+        return draco::GeometryAttribute::GENERIC;
+    }
+    return draco::GeometryAttribute::INVALID;
+}
+} // namespace
+
+CompressedMesh compressMesh(
+        const Qt3DRender::QGeometry &geometry,
+        const GLTF2ExportConfiguration &conf)
+{
+    std::unique_ptr<draco::EncoderBuffer> compressBuffer(new draco::EncoderBuffer);
+    std::vector<std::pair<QString, int>> attributes;
+    draco::Mesh dracoMesh;
+
+    // Convert Qt3D geometry to draco mesh
+    const auto geometryAttributes = geometry.attributes();
+    const Qt3DRender::QAttribute *indicesAttribute = nullptr;
+
+    for (const auto *attribute : geometryAttributes) {
+        if (attribute->name() == Qt3DRender::QAttribute::defaultPositionAttributeName()) {
+            dracoMesh.set_num_points(attribute->count());
+            break;
+        }
+    }
+
+    for (const auto *attribute : geometryAttributes) {
+        if (attribute->attributeType() == Qt3DRender::QAttribute::IndexAttribute) {
+            indicesAttribute = attribute;
+            continue;
+        }
+
+        switch (attribute->vertexBaseType()) {
+        case Qt3DRender::QAttribute::VertexBaseType::Float: {
+            if(!compressAttribute<float>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::Byte: {
+            static_assert(CHAR_MIN < 0, "This code only works on platforms with signed char");
+            if(!compressAttribute<char>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedByte: {
+            if(!compressAttribute<unsigned char>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::Short: {
+            if(!compressAttribute<short>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedShort: {
+            if(!compressAttribute<unsigned short>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::Int: {
+            if(!compressAttribute<int>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedInt: {
+            if(!compressAttribute<unsigned int>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::Double: {
+            if(!compressAttribute<double>(*attribute, dracoMesh, attributes))
+                return {};
+            break;
+        }
+        case Qt3DRender::QAttribute::VertexBaseType::HalfFloat:
+        default:
+            qCWarning(dracocompressor) << "Warning: skipping attribute" << attribute->name() << "of unhandled type" << attribute->vertexBaseType();
+            break;
+        }
+    }
+
+    if (indicesAttribute != nullptr) {
+        switch (indicesAttribute->vertexBaseType()) {
+        case Qt3DRender::QAttribute::VertexBaseType::Byte:
+            addFacesToMesh(createIndicesVectorFromAttribute<char>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedByte:
+            addFacesToMesh(createIndicesVectorFromAttribute<unsigned char>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::Short:
+            addFacesToMesh(createIndicesVectorFromAttribute<short>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedShort:
+            addFacesToMesh(createIndicesVectorFromAttribute<unsigned short>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::Int:
+            addFacesToMesh(createIndicesVectorFromAttribute<int>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedInt:
+            addFacesToMesh(createIndicesVectorFromAttribute<unsigned int>(*indicesAttribute), dracoMesh);
+            break;
+        default:
+            break;
+        }
+    }
+
+    draco::Encoder encoder;
+
+    auto setQuantization = [&](GLTF2ExportConfiguration::MeshAttribute attr) {
+        int q = conf.attributeQuantizationLevel(attr);
+        if (q > 0)
+            encoder.SetAttributeQuantization(dracoAttribute(attr), q);
+    };
+    setQuantization(GLTF2ExportConfiguration::Position);
+    setQuantization(GLTF2ExportConfiguration::Normal);
+    setQuantization(GLTF2ExportConfiguration::Color);
+    setQuantization(GLTF2ExportConfiguration::TextureCoordinate);
+    setQuantization(GLTF2ExportConfiguration::Generic);
+
+    encoder.SetSpeedOptions(conf.meshEncodingSpeed(), conf.meshDecodingSpeed());
+
+    encoder.SetTrackEncodedProperties(true);
+    const auto status = encoder.EncodeMeshToBuffer(dracoMesh, compressBuffer.get());
+    if (!status.ok()) {
+        qCWarning(dracocompressor) << status.code() << status.error_msg();
+        return {};
+    }
+
+    return { std::move(compressBuffer), attributes };
+}
+
+} // namespace DracoCompression
+} // namespace Kuesa
+QT_END_NAMESPACE
--- a/src/core/gltf2exporter/dracocompressor_p.h
+++ b/src/core/gltf2exporter/dracocompressor_p.h
@@ -1,0 +1,69 @@
+/*
+    dracocompressor_p.h
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Juan Jose Casafranca <juan.casafranca@kdab.com>
+
+    Licensees holding valid proprietary KDAB Kuesa licenses may use this file in
+    accordance with the Kuesa Enterprise License Agreement provided with the Software in the
+    LICENSE.KUESA.ENTERPRISE file.
+
+    Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as
+    published by the Free Software Foundation, either version 3 of the
+    License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef KUESA_GLTF2EXPORTER_DRACOCOMPRESSOR_P_H
+#define KUESA_GLTF2EXPORTER_DRACOCOMPRESSOR_P_H
+
+//
+//  NOTICE
+//  ------
+//
+// We mean it: this file is not part of the public API and could be
+// modified without notice
+//
+
+#include <draco/compression/encode.h>
+#include <memory>
+#include <vector>
+#include <QString>
+
+namespace draco {
+class EncoderBuffer;
+}
+
+QT_BEGIN_NAMESPACE
+
+namespace Qt3DRender {
+class QGeometry;
+}
+namespace Kuesa {
+class GLTF2ExportConfiguration;
+namespace DracoCompression {
+
+struct CompressedMesh {
+    std::unique_ptr<draco::EncoderBuffer> buffer;
+    std::vector<std::pair<QString, int>> attributes;
+};
+
+CompressedMesh compressMesh(const Qt3DRender::QGeometry &geometry, const GLTF2ExportConfiguration &);
+} // namespace DracoCompression
+} // namespace Kuesa
+
+QT_END_NAMESPACE
+
+#endif // KUESA_GLTF2EXPORTER_DRACOCOMPRESSOR_H
--- a/src/core/gltf2exporter/gltf2exporter_p.cpp
+++ b/src/core/gltf2exporter/gltf2exporter_p.cpp
@@ -1,0 +1,736 @@
+﻿/*
+    gltf2exporter_p.cpp
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Jean-Michaël Celerier <jean-michael.celerier@kdab.com>
+
+    Licensees holding valid proprietary KDAB Kuesa licenses may use this file in
+    accordance with the Kuesa Enterprise License Agreement provided with the Software in the
+    LICENSE.KUESA.ENTERPRISE file.
+
+    Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as
+    published by the Free Software Foundation, either version 3 of the
+    License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <draco/compression/encode.h>
+#if defined(KUESA_DRACO_COMPRESSION)
+#include "dracocompressor_p.h"
+#endif
+
+#include <QFile>
+#include <SceneEntity>
+#include "gltf2exporter_p.h"
+#include "gltf2context_p.h"
+#include "gltf2keys_p.h"
+#include "gltf2importer.h"
+#include "kuesa_p.h"
+#include <set>
+#include <Kuesa/private/gltf2context_p.h>
+
+QT_BEGIN_NAMESPACE
+
+namespace Qt3DRender {
+class QGeometryRenderer;
+class QGeometry;
+} // namespace Qt3DRender
+
+namespace Kuesa {
+namespace {
+QString generateUniqueFilename(const QDir& dir, QString filename)
+{
+    QFileInfo fi(filename);
+    auto basename = fi.baseName();
+    auto ext = fi.completeSuffix();
+    int num = 1;
+    while(dir.exists(filename))
+    {
+        filename = basename + "-" + QString::number(num) + "." + ext;
+    }
+    return filename;
+}
+}
+
+void GLTF2ExportConfiguration::setMeshEncodingSpeed(int speed)
+{
+    if (speed >= 0 && speed <= 10)
+        m_encodingSpeed = speed;
+}
+
+void GLTF2ExportConfiguration::setMeshDecodingSpeed(int speed)
+{
+    if (speed >= 0 && speed <= 10)
+        m_decodingSpeed = speed;
+}
+
+int GLTF2ExportConfiguration::meshEncodingSpeed() const
+{
+    return m_encodingSpeed;
+}
+
+int GLTF2ExportConfiguration::meshDecodingSpeed() const
+{
+    return m_decodingSpeed;
+}
+
+void GLTF2ExportConfiguration::setAttributeQuantizationLevel(
+        GLTF2ExportConfiguration::MeshAttribute attribute,
+        int level)
+{
+    if (level >= 0 && level <= 10)
+        m_quantization[attribute] = level;
+}
+
+int GLTF2ExportConfiguration::attributeQuantizationLevel(
+        GLTF2ExportConfiguration::MeshAttribute attribute) const
+{
+    auto it = m_quantization.find(attribute);
+    if (it != m_quantization.end()) {
+        return it.value();
+    }
+    return 0;
+}
+
+void GLTF2ExportConfiguration::setMeshCompressionEnabled(bool enabled)
+{
+    m_meshCompression = enabled;
+}
+
+bool GLTF2ExportConfiguration::meshCompressionEnabled() const
+{
+    return m_meshCompression;
+}
+
+namespace {
+#if defined(KUESA_DRACO_COMPRESSION)
+class GLTF2DracoCompressor
+{
+public:
+    GLTF2DracoCompressor(
+            QDir source,
+            QDir destination,
+            const QJsonObject &rootObject,
+            const GLTF2ExportConfiguration &conf,
+            const GLTF2Import::GLTF2Context &context)
+        : m_root(rootObject)
+        , m_buffers(rootObject[GLTF2Import::KEY_BUFFERS].toArray())
+        , m_bufferViews(rootObject[GLTF2Import::KEY_BUFFERVIEWS].toArray())
+        , m_accessors(rootObject[GLTF2Import::KEY_ACCESSORS].toArray())
+        , m_meshes(rootObject[GLTF2Import::KEY_MESHES].toArray())
+        , m_images(rootObject[GLTF2Import::KEY_IMAGES].toArray())
+        , m_compressedBufferIndex(m_buffers.size()) // Indice of the added buffer
+        , m_newBufferViewIndex(m_bufferViews.size()) // Indice of where the bufferView are added
+        , m_basePath(source)
+        , m_destination(destination)
+        , m_conf(conf)
+        , m_context(context)
+    {
+    }
+
+    QJsonObject compress()
+    {
+        // https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md
+
+        // 1. Compress all the meshes
+        for (int i = 0, n = m_context.meshesCount(); i < n; i++) {
+            const auto mesh_json = m_meshes[i].toObject();
+            m_meshes[i] = compressMesh(mesh_json, i);
+        }
+
+        if (m_compressedBuffer.isEmpty()) {
+            return m_root; // Nothing changed
+        }
+
+        // 2. Save all the compressed data in a monolithic buffer and add it to the buffer list
+        {
+
+            const auto defaultCompressedBufferFile = generateUniqueFilename(m_basePath, QStringLiteral("compressedBuffer.bin"));
+
+            // TODO maybe warn the user of unused assets ?
+            QFile compressedBufferFile(m_destination.filePath({ defaultCompressedBufferFile }));
+            if (compressedBufferFile.open(QIODevice::WriteOnly)) {
+                compressedBufferFile.write(m_compressedBuffer);
+                compressedBufferFile.close();
+            } else {
+                m_errors << QStringLiteral("Could not open %1 for writing.").arg(compressedBufferFile.fileName());
+                return {};
+            }
+
+            QJsonObject compressedBufferObject;
+            compressedBufferObject[GLTF2Import::KEY_BYTELENGTH] = m_compressedBuffer.size();
+            compressedBufferObject[GLTF2Import::KEY_URI] = defaultCompressedBufferFile;
+            m_buffers.push_back(compressedBufferObject);
+        }
+
+        // 3. Adjust the accessors: they don't need to point to a bufferView anymore since it's the Draco extension which is aware of them
+        // TODO what happens if an accessor was referred by a mesh but also something else ?
+        std::set<int> bufferViews_to_clean = cleanAccessors(m_accessors, m_accessorsToClean);
+
+        // 4. Remove the data from the buffers and adjust the remaining bufferViews
+        // We have to update *all* the indices everywhere - not only for meshes.
+        removeBufferViews(bufferViews_to_clean);
+
+        // 5. Finalize the JSON
+        auto replace_json = [&](QJsonObject &m_root, const QLatin1String &k, QJsonArray &arr) {
+            auto it = m_root.find(k);
+            if (it != m_root.end() && arr.empty())
+                m_root.erase(it);
+            else if (!arr.empty())
+                *it = std::move(arr);
+        };
+        replace_json(m_root, GLTF2Import::KEY_BUFFERS, m_buffers);
+        replace_json(m_root, GLTF2Import::KEY_BUFFERVIEWS, m_bufferViews);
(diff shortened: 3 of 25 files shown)


jcelerier:
@mkrus it's on top of current master now

mkrus:
guess this needs rebased on master once the parser context changes are merged

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -337,6 +345,42 @@ void MainWindow::openFile()
     setFilePath(filePath);
 }
 
+void MainWindow::compressFile()
```
mkrus: need error reporting if compressing fails, write files fails, etc

tests/auto/gltfexporter/tst_gltfexporter.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+    tst_gltfparser.cpp
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Mike Krus <mike.krus@kdab.com>
```
mkrus: not me :)

tests/auto/gltfexporter/gltfexporter.pro
```diff
@@ -0,0 +1,39 @@
+# gltfparser.pro
+#
+# This file is part of Kuesa.
+#
+# Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+# Author: Mike Krus <mike.krus@kdab.com>
```
mkrus: not me :)

src/core/gltf2importer/dracocompressor_p.h
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+#include <stdexcept>
+#include <QString>
+
+namespace Qt3DRender {
+class QGeometry;
+}
+namespace Kuesa {
+namespace Compression {
+struct DracoError : std::runtime_error {
```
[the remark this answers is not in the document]
jcelerier: done

src/core/gltf2importer/gltf2importer.h
```diff
@@ -45,6 +45,7 @@ class KUESASHARED_EXPORT GLTF2Importer : public Qt3DCore::QNode
     Q_PROPERTY(QUrl source READ source WRITE setSource NOTIFY sourceChanged)
     Q_PROPERTY(Kuesa::GLTF2Importer::Status status READ status NOTIFY statusChanged)
     Q_PROPERTY(Kuesa::SceneEntity *sceneEntity READ sceneEntity WRITE setSceneEntity NOTIFY sceneEntityChanged)
+    Q_PROPERTY(Kuesa::GLTF2Context *context READ context)
```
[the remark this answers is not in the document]
jcelerier: done

src/quick/imports/kuesaplugin.cpp
```diff
@@ -73,6 +74,7 @@ void KuesaPlugin::registerTypes(const char *uri)
 
     // Misc
     qmlRegisterType<Kuesa::GLTF2Importer>(uri, 1, 0, "GLTF2Importer");
+    qmlRegisterType<Kuesa::GLTF2Exporter>(uri, 1, 0, "GLTF2Exporter");
```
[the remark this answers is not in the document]
jcelerier: done (we discussed this a bit with @lemirep  IIRC)

tests/auto/gltfexporter/tst_gltfexporter.cpp
jcelerier: done

tests/auto/gltfexporter/gltfexporter.pro
jcelerier: done

tools/assetpipelineeditor/mainwindow.ui
```diff
@@ -183,20 +186,25 @@
...
   <action name="actionSettings">
    <property name="text">
-    <string>Settings</string>
+    <string>&amp;Settings</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+,</string>
    </property>
   </action>
+  <action name="actionSave">
```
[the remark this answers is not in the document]
jcelerier: ok, indeed save is not ideal especially considering that this is a fairly destructive operation - currently I save in a subfolder named "compressed". there could also be the possibility to set the compression ratio - draco seems to be tweakable per-attribute.

mkrus:
it's fine to the tool to access the private classes, they will all be release together

tests/auto/gltfexporter/tst_gltfexporter.cpp
```diff
@@ -4,7 +4,7 @@
     This file is part of Kuesa.
 
     Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
```
mkrus: 2019

tests/auto/gltfexporter/gltfexporter.pro
```diff
@@ -3,7 +3,7 @@
 # This file is part of Kuesa.
 #
 # Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
```
mkrus: 2019

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,275 @@
+#include "dracocompressor_p.h"
```
mkrus: missing copyright header

jcelerier:
I rebased on master and added a dialog that will allow to set-up the export options, hope the design is correct (the bottom part reports the errors that may happen during compression) : 
dialog

mkrus:
> I rebased on master and added a dialog that will allow to set-up the export options, hope the design is correct (the bottom part reports the errors that may happen during compression) :
> dialog

missing changes to `assetpipelineeditor.pro` to include new files

jcelerier:
Added the changes, sorry !

jcelerier:
> the OS will prompt you.

So, the only thing that may be dangerous is that you prompt the file chooser for a .gltf file, but it's not only the .gltf which will change but also the assets which will be stripped of data (and it cannot even be recovered since draco is lossy).
e.g. if for instance you save in the same folder but with a different name, you won't get prompted by the OS but the assets will silently change.

jcelerier:
I rebased on master, since GLTFContext is back to private and is needed for the export I added the exporter as a friend of the importer so that it can go and fetch it without having GLTFContext leaking in the public api. (if there are any better ways to do this please tell me !)

jcelerier:
Managed to fix the compression crash ! e.g. the dodgeViper does not crash anymore while compressing (it doesn't reload though, investigating.). Much more examples of the GLTF2 tests are able to compress and reload, e.g. here's a fairly quantized duck : 
ducky

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -0,0 +1,372 @@
...
+            {
+                QJsonObject draco_ext;
+                draco_ext[GLTF2Import::KEY_BUFFERVIEW] = newBufferViewIndex;
+                {
+                    QJsonObject draco_ext_attr;
+                    for (const auto &[name, idx] : compressed.attributes) {
+
+                        // TODO the names aren't the correct ones, how to get them ?
+                        // e.g. vertexNormal instead of NORMAL
+                        // for the sake of testing, this dirty hack :
```
lemirep: QAttribute::defaultPositionAttributeName, QAttribute::defaultNormalAttributeName, QAttribute::defaultColorAttributeName, QAttribute::defaultTextureCoordinateAttributeName, QAttribute::defaultTextureCoordinate1AttributeName, QAttribute::defaultTextureCoordinate2AttributeName,
QAttribute::defaultTangentAttributeName, QAttribute::defaultJointIndicesAttributeName, QAttribute::defaultJointWeightsAttributeName

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -235,23 +244,238 @@ class GLTF2DracoCompressor
...
 
+    void removeBufferViews(const std::set<int> &indices)
+    {
+        // First compute the new indicides of the buffer views
+        std::map<int, int> bufferViewIndex;
+        int currentOffset = 0;
+        for (int i = 0, n = m_bufferViews.size(); i < n; i++) {
+            if (indices.count(i) != 0) {
+                currentOffset++;
+            }
```
lemirep: ++currentOffset; {} not needed

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -235,23 +244,238 @@ class GLTF2DracoCompressor
...
+            bufferViewIndex[i] = i - currentOffset;
+        }
+
+        // Fix them in the accessors
+        // TODO sparse accessors
+        for (int i = 0; i < m_accessors.size(); i++) {
+            auto acc = m_accessors[i].toObject();
+
+            auto it = acc.find(GLTF2Import::KEY_BUFFERVIEW);
```
lemirep: Probably nicer to use const QJsonValue v = acc.value(KEY_BUFFERVIEW); if (!v.isUndefined()) ...

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -235,23 +244,238 @@ class GLTF2DracoCompressor
...
+        }
+
+        // Fix them in the meshes
+        for (int i = 0; i < m_meshes.size(); i++) {
+            auto mesh = m_meshes[i].toObject();
+            auto primitives = mesh[GLTF2Import::KEY_PRIMITIVES].toArray();
+            for (int p = 0; p < primitives.size(); p++) {
+                auto primitive = primitives[p].toObject();
+
+                auto ext_it = primitive.find(GLTF2Import::KEY_EXTENSIONS);
```
lemirep: I'd go for the QJsonValue here as well as it improves readability, the iterator only makes sense to avoid multiple lookups or to delete the entry.

src/core/gltf2importer/dracocompressor_p.cpp
```diff
@@ -149,11 +157,15 @@ QVector<T> createIndicesVectorFromAttribute(const Qt3DRender::QAttribute &attrib
     indicesVector.resize(attribute.count());
 
     if (byteStride == 0)
+    {
```
lemirep: We usually keep the opening { on the same line as the if statement

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -403,16 +392,19 @@ class GLTF2DracoCompressor
             QJsonObject buf = m_buffers[buffer.first].toObject();
 
             auto uri = buf[GLTF2Import::KEY_URI].toString();
-            QFile f(m_basePath.absoluteFilePath(uri));
-            if (!f.open(QIODevice::ReadWrite))
+            QFile source_f(m_basePath.absoluteFilePath(uri));
+            if (!source_f.open(QIODevice::ReadOnly))
+            {
```
lemirep: { should be on same line as if

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -546,16 +543,37 @@ void GLTF2Exporter::setContext(GLTF2Context *context)
...
 #if defined(KUESA_DRACO_COMPRESSION)
-    GLTF2DracoCompressor compressor(rootObject, gltfFileDir, ctx);
-    rootObject = compressor.compress(gltfFileDir);
+    GLTF2DracoCompressor compressor(rootObject, source, target, ctx);
+    rootObject = compressor.compress();
+    if(rootObject.empty())
+        return {};
 #endif
 
+    if(source != target)
```
lemirep: missing space after if, { should be on same line

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -546,16 +543,37 @@ void GLTF2Exporter::setContext(GLTF2Context *context)
...
+    {
+        // Copy images
+        const auto images = rootObject[GLTF2Import::KEY_IMAGES].toArray();
+        for (const auto& val : images)
+        {
```
lemirep: here too

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -546,16 +543,37 @@ void GLTF2Exporter::setContext(GLTF2Context *context)
...
+            const auto& image = val.toObject();
+            auto uri_it = image.find(GLTF2Import::KEY_URI);
+            if(uri_it != image.end())
+            {
```
lemirep: and here

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -546,16 +543,37 @@ void GLTF2Exporter::setContext(GLTF2Context *context)
...
+                QFileInfo destFile(target, uri_it->toString());
```
lemirep: const QFileInfo

src/core/gltf2importer/gltf2exporter.h
```diff
@@ -61,7 +61,10 @@ public Q_SLOTS:
     void setContext(GLTF2Context *context);
     void setScene(SceneEntity *scene);
 
-    QJsonObject compress(QDir folder, QJsonObject rootObject);
+    QJsonObject compress(
+            QDir sourceFolder,
+            QDir targetFolder,
+            QJsonObject rootObject);
```
lemirep: const QJsonObject &

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -362,12 +362,18 @@ void MainWindow::compressFile()
         Kuesa::GLTF2Exporter exporter;
         exporter.setContext(importer->context());
         exporter.setScene(m_entity);
-        rootObject = exporter.compress(QFileInfo(gltfFile).dir(), rootObject);
+
+        auto srcdir = QFileInfo(gltfFile).dir();
+        srcdir.mkdir("compressed");
+
+        auto targetdir = srcdir;
```
lemirep: const auto targetdir = srcdir.cd("compressed");

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -362,12 +362,18 @@ void MainWindow::compressFile()
...
+        targetdir.cd("compressed");
+
+        rootObject = exporter.compress(srcdir, targetdir, rootObject);
         if(rootObject.empty())
```
lemirep: space after if

tests/auto/gltfexporter/gltfexporter.pro
lemirep: I don't think Mike actually wrote this :)

tests/auto/gltfexporter/tst_gltfexporter.cpp
lemirep: Really nice of you :)

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -99,18 +99,25 @@ class GLTF2DracoCompressor
...
 
         // 5. Finalize the JSON
-        m_root[GLTF2Import::KEY_BUFFERS] = std::move(m_buffers);
-        m_root[GLTF2Import::KEY_BUFFERVIEWS] = std::move(m_bufferViews);
-        m_root[GLTF2Import::KEY_ACCESSORS] = std::move(m_accessors);
-        m_root[GLTF2Import::KEY_MESHES] = std::move(m_meshes);
-        m_root[GLTF2Import::KEY_IMAGES] = std::move(m_images);
+        auto replace_json = [&] (QJsonObject& m_root, const QLatin1String& k, QJsonArray& arr) {
+            auto it = m_root.find(k);
+            if(it != m_root.end() && arr.empty())
```
lemirep: spaces between if and (

src/core/gltf2importer/gltf2exporter.cpp
```diff
@@ -559,18 +568,22 @@ QJsonObject GLTF2Exporter::compress(
     if(source != target)
     {
         // Copy images
-        const auto images = rootObject[GLTF2Import::KEY_IMAGES].toArray();
-        for (const auto& val : images)
+        auto images_it = rootObject.find(GLTF2Import::KEY_IMAGES);
+        if(images_it != rootObject.end())
         {
```
lemirep: { on if line

tests/auto/gltfexporter/tst_gltfexporter.cpp
```diff
@@ -33,43 +33,112 @@
...
 #include <Kuesa/private/kuesa_utils_p.h>
-
 using namespace Kuesa;
 using namespace GLTF2Import;
 
 class tst_GLTFExporter : public QObject
 {
     Q_OBJECT
 
+    QDir setupTestFolder()
```
lemirep: If would even go for a free standing function in an anonymous namespace

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -45,9 +38,24 @@ void ExportDialog::on_save()
 {
     ui->errorLog->setPlainText("");
 
-    auto rootObject = m_exporter.saveInFolder(
-            QFileInfo(m_originalFile).dir(),
-            QFileInfo(m_targetFile).dir());
+    m_targetFile = QFileDialog::getSaveFileName(this, tr("Select a folder to save to"), m_targetFile, tr("glTF 2.0 file (*.gltf)"));
+
+    QDir orig_dir = QFileInfo(m_originalFile).dir();
```
lemirep: const ?

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -45,9 +38,24 @@ void ExportDialog::on_save()
...
+    QDir target_dir = QFileInfo(m_targetFile).dir();
```
lemirep: const ?

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -45,9 +38,24 @@ void ExportDialog::on_save()
...
+    if(orig_dir.absolutePath() == target_dir.absolutePath())
+    {
```
lemirep: if style

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -45,9 +38,24 @@ void ExportDialog::on_save()
...
+        int res = QMessageBox::warning(
+                    this, tr("Destructive operation"),
+                    tr("Saving in the original folder will overwrite assets."),
+                    QMessageBox::Ok | QMessageBox::Cancel,
+                    QMessageBox::Cancel);
+        if(res != QMessageBox::Ok) {
```
lemirep: space after if

src/core/gltf2exporter/gltf2exporter.h
```diff
@@ -33,13 +33,50 @@
...
 QT_BEGIN_NAMESPACE
 
 namespace Kuesa {
 class GLTF2Context;
 class SceneEntity;
+
+class KUESASHARED_EXPORT GLTF2ExportConfiguration
+{
+public:
+    GLTF2ExportConfiguration();
```
lemirep: why not have = default here directly?

src/core/gltf2exporter/gltf2exporter.h
```diff
@@ -33,13 +33,50 @@
...
+    ~GLTF2ExportConfiguration();
+    GLTF2ExportConfiguration(const GLTF2ExportConfiguration &other);
+    GLTF2ExportConfiguration(GLTF2ExportConfiguration &&other);
+    GLTF2ExportConfiguration &operator=(const GLTF2ExportConfiguration &other);
+    GLTF2ExportConfiguration &operator=(GLTF2ExportConfiguration &&other);
+
```
lemirep: Do you even need to define all of these if all are the default?

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -187,12 +187,12 @@ QVector<T> createIndicesVectorFromAttribute(const Qt3DRender::QAttribute &attrib
     if (byteStride == 0)
     {
         for (int i = 0, nbIndices = indicesVector.size(); i < nbIndices; ++i)
-            indicesVector[i] = static_cast<T>(data.data()[i * sizeof(T)]);
```
lemirep: why not do const T *attributeData = static_cast<T *>(data.constData()); and in the loop indicesVector[i] = attributeData[i]; ?

src/core/gltf2importer/gltf2exporter.cpp
jcelerier: I should maybe squash my commits so that there is only the last version since there has been a bunch of back-and-forth,  what do you think ?

tests/auto/gltfexporter/gltfexporter.pro
jcelerier: likewise, in the latest commit it has already been fixed (https://github.com/KDAB/kuesa/pull/16/files#diff-1f8bf50859e4b422c66a957bfb568cafR6)

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+template<typename T>
+QByteArray packedDataInBuffer(const QByteArray &inputBuffer,
+                              int vertexSize,
+                              int offset,
+                              int stride,
+                              int count)
+{
+    QByteArray outputBuffer;
+    outputBuffer.resize(count * vertexSize * sizeof(T));
```
lemirep: vertexSize * sizeof(T) could probably cached into a variable since it's reused in several places

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,312 @@
...
+QVector<T> createIndicesVectorFromAttribute(const Qt3DRender::QAttribute &attribute)
+{
+    const auto byteStride = attribute.byteStride();
+    const auto byteOffset = attribute.byteOffset();
+    const auto data = attribute.buffer()->data().mid(byteOffset);
+
+    QVector<T> indicesVector;
+    indicesVector.resize(attribute.count());
+
+    if (byteStride == 0)
```
lemirep: if / else style

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+template<typename T>
+QVector<T> createIndicesVectorFromAttribute(const Qt3DRender::QAttribute &attribute)
+{
+    const auto byteStride = attribute.byteStride();
+    const auto byteOffset = attribute.byteOffset();
+    const auto data = attribute.buffer()->data().mid(byteOffset);
+
+    QVector<T> indicesVector;
+    indicesVector.resize(attribute.count());
```
lemirep: could be written as QVector<T> indicesVector(attribute.count()); directly

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,312 @@
...
+            dracoMesh.set_num_points(attribute->count());
+    }
+
+    for (const auto attribute : geometryAttributes) {
+        if (attribute->attributeType() == Qt3DRender::QAttribute::IndexAttribute) {
+            indicesAttribute = attribute;
+            continue;
+        }
+
+        qDebug() << "compressing: " << attribute->name() << attribute->vertexBaseType() << attribute->attributeType();
```
lemirep: maybe use categorized logging instead?

src/core/gltf2exporter/dracocompressor_p.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QString>
+
+namespace Qt3DRender {
+class QGeometry;
+}
+namespace Kuesa {
+class GLTF2ExportConfiguration;
+namespace Compression {
+
+struct CompressedMesh {
```
lemirep: { should be on new line for structs and classes I believe

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -0,0 +1,88 @@
...
+void ExportDialog::on_save()
+{
+    ui->errorLog->setPlainText("");
+
+    m_targetFile = QFileDialog::getSaveFileName(this, tr("Select a folder to save to"), m_targetFile, tr("glTF 2.0 file (*.gltf)"));
+
+    QDir orig_dir = QFileInfo(m_originalFile).dir();
+    QDir target_dir = QFileInfo(m_targetFile).dir();
+    if (orig_dir.absolutePath() == target_dir.absolutePath()) {
+        int res = QMessageBox::warning(
```
lemirep: const int res

src/core/gltf2importer/gltf2exporter.cpp
lemirep: Actually that's fine. I found I could use the Files Changed tab to do the review which gives me the same as if all commits had been squashed. Please ignore comments which aren't relevant any more.

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: done (also is there a way to do it without initializing values ? considering that they will be set just afterwards)

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: done (I changed this part so that there isn't any if / else since we can precompute what the actual stride will be)

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: I've removed this qdebug, it was only to help me fix a bug - what kind of useful logging could be added instead ?

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,312 @@
...
+            addFacesToMesh(createIndicesVectorFromAttribute<int>(*indicesAttribute), dracoMesh);
+            break;
+        case Qt3DRender::QAttribute::VertexBaseType::UnsignedInt:
+            addFacesToMesh(createIndicesVectorFromAttribute<unsigned int>(*indicesAttribute), dracoMesh);
+            break;
+        }
+    }
+
+    auto bb = dracoMesh.ComputeBoundingBox();
+    qDebug() << bb.max_point()[0]
```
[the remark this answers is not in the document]
jcelerier: no, removed

src/core/gltf2exporter/dracocompressor_p.h
jcelerier: hmm, when I run clang-format on that file (with `clang-format -i dracocompressor_p.h`)  it puts the brace here

tests/auto/gltfexporter/tst_gltfexporter.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: I've added error reporting in the export gui, is it enough ?

tools/assetpipelineeditor/mainwindow.ui
```diff
@@ -24,14 +24,15 @@
...
      <y>0</y>
      <width>1400</width>
-     <height>22</height>
+     <height>23</height>
     </rect>
    </property>
    <widget class="QMenu" name="menuFile">
     <property name="title">
-     <string>File</string>
+     <string>Fi&amp;le</string>
```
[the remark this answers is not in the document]
jcelerier: done

tools/assetpipelineeditor/mainwindow.ui
```diff
@@ -61,7 +62,7 @@
   </widget>
   <widget class="QDockWidget" name="collectionBrowserDockWidget">
    <property name="windowTitle">
-    <string>Collection Browser</string>
+    <string>&amp;Collection Browser</string>
```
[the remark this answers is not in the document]
jcelerier: done for this, but where is the Camera ? I don't find it in the mainwindow form

tools/assetpipelineeditor/mainwindow.ui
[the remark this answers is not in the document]
jcelerier: done (changed to Export)

tools/assetpipelineeditor/mainwindow.ui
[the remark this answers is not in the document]
jcelerier: found it

tools/assetpipelineeditor/dracocompressor.cpp
```diff
@@ -0,0 +1,166 @@
...
+            indicesAttribute = attribute;
+
+        const auto dracoDataType = dataTypeFromVertexBaseType(attribute->vertexBaseType());
+        if (dracoAttributeType == draco::GeometryAttribute::Type::INVALID ||
+                        dracoDataType == draco::DataType::DT_INVALID)
+            continue;
+
+        const auto attributeBuffer = attribute->buffer();
+        if (attributeBufferToDataBuffer.find(attributeBuffer) == std::end(attributeBufferToDataBuffer))
+            attributeBufferToDataBuffer[attributeBuffer] = createDracoBufferFromQBuffer(*attributeBuffer);
```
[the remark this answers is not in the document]
jcelerier: this code has since changed

tools/assetpipelineeditor/dracocompressor.cpp
```diff
@@ -0,0 +1,166 @@
...
+            indicesVector[i] = static_cast<T>(data.data()[i*byteStride]);
+
+    return indicesVector;
+}
+
+template <typename T>
+void addFacesToMesh(const QVector<T> &indices, draco::Mesh &dracoMesh)
+{
+    for (int i, nbIndices = indices.size()/3; i < nbIndices; ++i)
+        dracoMesh.AddFace({static_cast<draco::IndexType<unsigned int, draco::PointIndex_tag_type_>>(indices[3*i+0]),
```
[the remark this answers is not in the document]
jcelerier: it seems that it does, this has been fixed

tools/assetpipelineeditor/dracocompressor.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <memory>
+#include <draco/compression/encode.h>
+
+namespace Qt3DRender
+{
+class QGeometry;
+};
+
+std::unique_ptr<draco::EncoderBuffer> compressMesh(const Qt3DRender::QGeometry &geometry);
```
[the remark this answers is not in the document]
jcelerier: I had named it "Compression" in a further commit, do you want me to change it to DracoCompressor ?

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

src/core/gltf2exporter/gltf2exporter.h
jcelerier: actually no, I originally took inspiration from other Qt classes but they are not needed at all - removed.

src/core/gltf2importer/gltf2exporter.h
jcelerier: I also pass the QDir as const& now, is it ok ?

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: does not exist anymore

lemirep:
Kuesa: WARNING: kuesa/src/core/gltf2exporter/dracocompressor_p.h does not have the "We mean it." warning
Kuesa: WARNING: kuesa/src/core/gltf2exporter/gltf2exporter_p.h does not have the "We mean it." warning

src/core/gltf2exporter/gltf2exporter_p.h
```diff
@@ -0,0 +1,115 @@
...
+class GLTF2Importer;
+namespace GLTF2Import {
+class GLTF2Context;
+}
+class SceneEntity;
+
+class KUESASHARED_EXPORT GLTF2ExportConfiguration
+{
+public:
+    enum MeshAttribute { Position,
```
er-vin: Please put Position on its own line

src/core/gltf2exporter/dracocompressor_p.h
er-vin: yes, structs and classes have different rules. structs are treated like enums regarding that point.

src/core/gltf2exporter/gltf2exporter_p.h
```diff
@@ -0,0 +1,115 @@
...
+public:
+    explicit GLTF2Exporter(QObject *parent = nullptr);
+
+    SceneEntity *scene() const;
+    QStringList errors() const;
+
+    GLTF2ExportConfiguration configuration() const;
+    void setConfiguration(const GLTF2ExportConfiguration &conf);
+
+    void setContext(GLTF2Import::GLTF2Context *context);
```
er-vin: Context is private API so it's unlikely user code will be able to put one here (or we force them to do use kuesa-private). Are we sure this is required?
Ditto for the following method, this context class really feels like an implementation detail leaking through API.

tests/auto/gltfexporter/tst_gltfexporter.cpp
```diff
@@ -0,0 +1,198 @@
+/*
+    tst_gltfparser.cpp
```
er-vin: Wrong filename?

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <QFileInfo>
+#include <QDir>
+#include <QJsonDocument>
+#include <QFileDialog>
+#include <QStyle>
+#include <QPushButton>
+#include <QMessageBox>
+
+ExportDialog::ExportDialog(
+        Kuesa::GLTF2Exporter &exporter,
```
er-vin: Parameters would be better on a single line.

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -0,0 +1,88 @@
...
+        const QString &path,
+        QWidget *parent)
+    : QDialog(parent), ui(new Ui::ExportDialog), m_exporter(exporter), m_originalFile(path), m_targetFile(m_originalFile)
```
er-vin: OTOH it'd be best to have one per line here.

tools/assetpipelineeditor/exportdialog.h
```diff
@@ -0,0 +1,33 @@
...
+    Q_OBJECT
+
+public:
+    explicit ExportDialog(Kuesa::GLTF2Exporter &exporter,
+                          const QString &originalPath,
+                          QWidget *parent = nullptr);
+    ~ExportDialog();
+
+private:
+    void on_save();
```
er-vin: Use camel case like everything else.

src/core/gltf2exporter/dracocompressor_p.cpp
lemirep: You could keep it but maybe log it into a dedicated category Kuesa::DracoCompressor or something?

src/core/gltf2exporter/dracocompressor_p.cpp
lemirep: I guess you could do QVector<T> indicesVector; indicesVector.reserve(attribute.count()); which should preallocate without initializing value. Then you need to use push_back to insert values though.

tools/assetpipelineeditor/dracocompressor.h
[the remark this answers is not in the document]
lemirep: That could be a good idea, I suspect we might have other Compression classes that deal with Textures or other unrelated assets later on.

tools/assetpipelineeditor/dracocompressor.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        const auto dracoBuffer = attributeBufferToDataBuffer[attributeBuffer];
+
+        auto meshAttribute = std::make_unique<draco::GeometryAttribute>();
+        meshAttribute->Init(dracoAttributeType,
+                            dracoBuffer,
+                            attribute->vertexSize(),
+                            dracoDataType,
+                            false,
+                            attribute->byteStride(),
```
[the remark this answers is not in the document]
jcelerier: done

tools/assetpipelineeditor/dracocompressor.h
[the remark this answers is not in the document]
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: hmm I think that it would be a bit less performant - let's keep it as ` QVector<T> indicesVector(static_cast<int>(attrCount))` :)

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.h
jcelerier: I added a destructor to that struct (in order to not require a draco header in that file) - should I keep it as a struct or change it to a class ?

src/core/gltf2exporter/gltf2exporter_p.h
jcelerier: hmm, it is "only" needed to access the structures defined in meshparser_p.h - I don't see other easy ways right now (short of re-parsing the original gltf inside the gltfexporter, recreating the qt3d nodes, and operating on those instead).

jcelerier: another solution would be to rewrite the dracocompressor file so that it isn't based on qt3d entities / attributes / etc. However this will break the day where the scene can be modified / combined / whatever, and since it is on the roadmap...

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: done

tests/auto/gltfexporter/tst_gltfexporter.cpp
jcelerier: done

jcelerier:
@er-vin @lemirep  I added my last changes in a new commit to simplify review for you, when it's done I can split them in per-independent-class commits if you want (compressor, exporter, ui ?). However it's fairly hard to divide in commits for the various KUE-387 tasks, because if they aren't all done together then the export will produce meaningless gltf files.

jcelerier:
Finally fixed the export of the DodgeViper - joints indices weren't exported properly.

src/core/gltf2exporter/gltf2exporter_p.h
jcelerier: done

src/core/gltf2exporter/gltf2exporter_p.h
mkrus: well GLTF2Exporter is also private so what's the issue?

src/core/gltf2importer/gltf2keys_p.h
```diff
@@ -0,0 +1,88 @@
+/*
+    gltf2keys_p.h
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
```
mkrus: 2019

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include <QFile>
+#include <QLoggingCategory>
+#include <QDataStream>
+#include <fstream>
+#include <iterator>
+#include <iostream>
+
+QT_BEGIN_NAMESPACE
+Q_LOGGING_CATEGORY(dracocompressor, "DracoCompressor")
+namespace Kuesa::DracoCompression {
```
mkrus: clang complains nested namespaces are a C++17 extension. We're targeting c++11

src/core/gltf2exporter/dracocompressor_p.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Qt3DRender {
+class QGeometry;
+}
+namespace Kuesa {
+class GLTF2ExportConfiguration;
+namespace DracoCompression {
+
+struct CompressedMesh {
+    ~CompressedMesh();
```
mkrus: why is the destructor needed? Adding it disables the generation of the default copy constructor which (with clang) leads to compile errors

src/core/gltf2exporter/gltf2exporter_p.cpp
```diff
@@ -0,0 +1,716 @@
...
+    }
+
+    // Compress a single primitive of a mesh
+    CompressedGLTFPrimitive compressPrimitive(
+            QJsonObject primitive_json, const Kuesa::GLTF2Import::Primitive &primitive) const
+    {
+        CompressedGLTFPrimitive compressed_primitive;
+
+        // Do the compression
+        const auto compressed = Kuesa::DracoCompression::compressMesh(*primitive.primitiveRenderer->geometry(), m_conf);
```
mkrus: this fails to compile with clang if the destructor is added to the `CompressedMesh` class (which disables the generation of the default copy constructor)

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,314 @@
...
+};
+
+template<>
+struct attribute_type_trait<float> {
+    static constexpr draco::DataType type = draco::DataType::DT_FLOAT32;
+};
+
+template<>
+struct attribute_type_trait<char> {
+    static constexpr draco::DataType type = draco::DataType::DT_INT8;
```
mkrus: clang warns that `type` is unused (as well as DT_UINT8, DT_INT16, DT_INT32, DT_UINT32, DT_FLOAT64).

src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -0,0 +1,314 @@
...
+    for (int i = 0, n = int(attrCount); i < n; ++i)
+        indicesVector[i] = *reinterpret_cast<const T *>(data.data() + i * stride);
+
+    return indicesVector;
+}
+
+template<typename T>
+void addFacesToMesh(const QVector<T> &indices, draco::Mesh &dracoMesh)
+{
+    using index_type = draco::IndexType<unsigned int, draco::PointIndex_tag_type_>;
```
mkrus: why this? Looks like `Face` is just an array of `PointIndex`?

tools/assetpipelineeditor/exportdialog.ui
```diff
@@ -0,0 +1,314 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mkrus: could you remove the enabled and make the `QGroupBox` checkable?

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -51,7 +52,14 @@
...
 #include <QClipboard>
 
+#include <QJsonDocument>
+#include <QJsonValue>
+#include <QJsonObject>
+#include <QJsonArray>
+#include <GLTF2Importer>
+
 #include <Kuesa/SceneEntity>
+#include <private/gltf2exporter_p.h>
```
mkrus: `Kuesa/private/...` I think is needed

tools/assetpipelineeditor/mainwindow.ui
```diff
@@ -131,7 +134,7 @@
   </widget>
   <widget class="QDockWidget" name="animationDockWidget">
    <property name="windowTitle">
-    <string>Animations</string>
+    <string>A&amp;nimations</string>
```
mkrus: is A used? would make more sense...

src/core/gltf2exporter/dracocompressor_p.h
mkrus: weird, if I remove the destructor it builds for me. Stupid compilers.

Don't mind on the fix, but if you add the destructor then you need also add the copy constructor (at least, maybe the regular constructor too) for clang to be happy

tools/assetpipelineeditor/exportdialog.ui
jcelerier: done

tools/assetpipelineeditor/mainwindow.ui
jcelerier: it's used for "About" but maybe that one is less relevant

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.cpp
jcelerier: done

src/core/gltf2importer/gltf2keys_p.h
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.h
jcelerier: > weird, if I remove the destructor it builds for me. Stupid compilers.

I just checked and this looks like it's a matter of language version used - this does not work for me on gcc with C++11 either

jcelerier: ah, the reason why it may have worked without destructor nor include, is because all the files that use dracocompressor_p.h also include the required draco header - but a new source file that would just do 

     #include "dracocompressor_p.h"
     void foo() { CompressedMesh mesh; } 

would fail to build

jcelerier:
hmm I don't know why Sean's commit got in the middle, gonna rebase this

jcelerier:
@er-vin do you want me to squash and recreate atomic commits for the various features from this ?

src/core/gltf2exporter/dracocompressor_p.h
jcelerier: done

src/core/gltf2exporter/dracocompressor_p.h
```diff
@@ -0,0 +1,69 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef KUESA_GLTF2EXPORTER_DRACOCOMPRESSOR_H
```
er-vin: Nitpick: should end with _P_H

src/core/gltf2exporter/dracocompressor_p.h
er-vin: I'd say class in such a case. But I don't see the aforementioned destructor anyway.

src/core/gltf2exporter/gltf2exporter_p.h
```diff
@@ -0,0 +1,125 @@
+/*
+    gltf2exporter.h
```
er-vin: gltf2exporter_p.h

src/core/gltf2exporter/gltf2exporter_p.h
```diff
@@ -0,0 +1,125 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef KUESA_GLTF2EXPORTER_GLTF2EXPORTER_H
```
er-vin: Should use _P_H

src/core/gltf2exporter/gltf2exporter_p.h
er-vin: Oh right, I was confused by the export macro... But indeed for some reason we don't have a KUESA_PRIVATE_EXPORT macro. Maybe that should be changed? (in a separate commit)

tests/auto/gltfexporter/tst_gltfexporter.cpp
```diff
@@ -0,0 +1,198 @@
...
+            exporter.setScene(&scene);
+            exporter.setConfiguration(configuration);
+
+            QFile asset_file(asset);
+            asset_file.open(QIODevice::ReadOnly);
+            new_asset = exporter.saveInFolder(
+                    QDir(ASSETS),
+                    tmp);
+        }
+    }
```
er-vin: Can't we at least check something to verify the compression occured? It's lacking some QCOMPARE or QVERIFY in there. :-)

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -0,0 +1,104 @@
...
+    const QDir target_dir = QFileInfo(m_targetFile).dir();
+    if (orig_dir.absolutePath() == target_dir.absolutePath()) {
+        const int res = QMessageBox::warning(
+                this, tr("Destructive operation"),
+                tr("Saving in the original folder will overwrite assets."),
+                QMessageBox::Ok | QMessageBox::Cancel,
+                QMessageBox::Cancel);
+        if (res != QMessageBox::Ok) {
+            ui->errorLog->setPlainText(tr("Nothing saved."));
+            return;
```
er-vin: Unsure since I couldn't try it yet (I lack draco), but isn't it going to close the dialog still when getting there? If yes this should be fixed.

tools/assetpipelineeditor/exportdialog.cpp
```diff
@@ -0,0 +1,104 @@
...
+ExportDialog::~ExportDialog()
+{
+    delete ui;
+}
+
+void ExportDialog::onSave()
+{
+    ui->errorLog->setPlainText("");
+
+    m_targetFile = QFileDialog::getSaveFileName(this, tr("Select a folder to save to"), m_targetFile, tr("glTF 2.0 file (*.gltf)"));
```
er-vin: We're not handling the case where the user clicked cancel and we got an empty string back. We should likely bail out early then. Would be even better if that was selected from within the original dialog instead of a file dialog on top of a dialog when clicking save.

tools/assetpipelineeditor/mainwindow.cpp
er-vin: I guess it'll do for now

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -337,6 +350,21 @@ void MainWindow::openFile()
     setFilePath(filePath);
 }
 
+void MainWindow::compressFile()
```
er-vin: Probably shouldn't be name compress since it'll be more generally about exporting.

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -164,6 +173,10 @@ void MainWindow::updateScene(Kuesa::SceneEntity *entity)
         model->update(entity);
     m_ui->collectionBrowser->expandAll();
     m_ui->actionReload->setEnabled(true);
+#if defined(KUESA_DRACO_COMPRESSION)
+    m_ui->actionExport->setEnabled(true);
```
er-vin: I'd go as far as completely hiding it from the GUI for now as long as it's useful only in the draco case. Will change once textures can be marked for post-processing too.

src/core/gltf2exporter/dracocompressor_p.h
jcelerier: done

src/core/gltf2exporter/gltf2exporter_p.h
jcelerier: done

src/core/gltf2exporter/gltf2exporter_p.h
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: hmm, on my machine (linux) the export dialog isn't closed after saving, but maybe this is something platform-specific - I will try on mac & windows

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: is the best way to go towards this is to make the ExportDialog inherit from QFileDialog and add the compression controls at the bottom ? (and is it ok if it's not the native file dialogs anymore in that case ?)

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: done

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: done

tools/assetpipelineeditor/mainwindow.cpp
jcelerier: done

tests/auto/gltfexporter/tst_gltfexporter.cpp
jcelerier: done

tools/assetpipelineeditor/exportdialog.cpp
jcelerier: also it used to be like this before, should I maybe revert to that ? https://user-images.githubusercontent.com/2772730/51386615-57104480-1b23-11e9-8095-e198d147c6b7.png

er-vin: Yes, I think I preferred that older approach indeed. Sorry for the rework incurred.

tools/assetpipelineeditor/exportdialog.cpp
er-vin: OK, just me being confused on save accepting the dialog or not. I was surely confused, feel free to ignore this.

src/core/gltf2exporter/gltf2exporter_p.h
er-vin: Now that it's introduced (thanks btw!), maybe it's time to start using it here?

tools/assetpipelineeditor/exportdialog.cpp
mkrus: old approach had the issue that it didn't warm the user about overwriting the original file

er-vin: Clearly it should warn of course, this is unrelated to how the GUI though.

jcelerier: mesh here's how it looks now - the icon on the lineedit opens the save dialog. Is it ok ? (it seemed clearer to me than the previous solution)

jcelerier:
just pushed with the changes requested - here's the diff attached if it can make it simpler to review (I also added a safeguard for the created name of the compressed binary file - if there is already one with this name in the source folder it will try to use another - compressedBuffer-1.bin, etc) : 
[out.txt](https://github.com/KDAB/kuesa/files/2859690/out.txt)

PR #20, opened 2019-01-15
More demo animations

mkrus:
adds animations to show of skybox/env map, opacity mask, changing colours.

Also add a text overlay indicating what is going on

### DIFF
--- a/examples/kuesa/car-scene/qml/main.qml
+++ b/examples/kuesa/car-scene/qml/main.qml
@@ -99,13 +99,6 @@ Item {
         }
     }
 
-    function resetControls() {
-        speedC.value = 0
-        openLeftDoorSwitch.checked = false
-        openRightDoorSwitch.checked = false
-        openHoodSwitch.checked = false
-    }
-
     // to detect mouse activity and stop animations
     MouseArea {
         id: mainMouseArea
@@ -532,10 +525,7 @@ Item {
         repeat: false
         running: true
         interval: 2 * 60 * 1000 // 2 minutes
-        onTriggered: {
-            console.log("Triggered")
-            idleAnimationRunning = true
-        }
+        onTriggered: idleAnimationRunning = true
     }
 
     function restartIdleTimer() {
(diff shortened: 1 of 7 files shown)


jjcasmar:
Looks nice, but, at least in my PC, when we deactivate opacity mask, screen turns black, and the skybox has to be changed. As in this case we are activating/deactivating the opacity mask automatically, Ill suggest to wait until we fix that bug before merging this.

jjcasmar:
It happens also when I enable/disable manually the transparency mask.

examples/kuesa/car-scene/qml/main.qml
```diff
@@ -588,8 +588,69 @@ Item {
             PauseAnimation { duration: 1000 }
         }
```
lemirep: Would it make sense to move this into a DemoIdleAnimation element? I feel this might unnecessarily clutter the main.qml when people are getting started with this example.

examples/kuesa/car-scene/qml/main.qml
```diff
@@ -665,4 +668,51 @@ Item {
         loops: Animation.Infinite
         running: idleAnimationRunning
     }
+
+    // Title Panel & Animation
```
lemirep: I'd move this in a separate file as well to keep main.qml from exploding. I'd even be tempted to move the whole side panel into a dedicated component so that main.qml is really about instantiating the 3D view and the various UI panels

examples/kuesa/car-scene/qml/main.qml
er-vin: Agreed.

examples/kuesa/car-scene/qml/main.qml
er-vin: +1

examples/kuesa/car-scene/qml/main.qml
mkrus: done

examples/kuesa/car-scene/qml/main.qml
mkrus: done, moved menu out of main, and animation also in separate file

lemirep:
It does work well. I did notice that when I or the animation switches the opacity mask then turns it off, once turned off, everything becomes black. Since it also happens without this set of patches I can only conclude that it's a separate issue, I'll file a bug report for that

PR #34, opened 2019-02-08
Fix incorrect export macro name

jcelerier:
Sorry for this


jcelerier:
yes, that's when I wanted to start using it in the other PR that I noticed the mistake :)

PR #37, opened 2019-02-12
Compute tangents if not provided

jjcasmar:
glTF 2.0 requires we compute tangents if necessary using MikkTSpace. 
We could only generate them if the material for a primitive has a normal map, but since we can change on runtime the material, I have chosen to compute them always, is possible. 

Probably still need some massage for making it faster.

### DIFF
--- a/src/core/gltf2importer/gltf2importer.pri
+++ b/src/core/gltf2importer/gltf2importer.pri
@@ -44,7 +44,8 @@ SOURCES += \
     $$PWD/sceneparser.cpp \
     $$PWD/materialparser.cpp \
     $$PWD/skinparser.cpp \
-    $$PWD/gltf2uri.cpp
+    $$PWD/gltf2uri.cpp \
+    $$PWD/meshparser_utils.cpp
 
 
 HEADERS += \
@@ -67,6 +68,7 @@ HEADERS += \
     $$PWD/skinparser_p.h \
     $$PWD/gltf2keys_p.h \
     $$PWD/gltf2uri_p.h \
+    $$PWD/meshparser_utils_p.h
 
 qtConfig(kuesa-draco) {
     DEFINES += KUESA_DRACO_COMPRESSION
--- a/src/core/gltf2importer/meshparser.cpp
+++ b/src/core/gltf2importer/meshparser.cpp
@@ -125,6 +125,7 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
     return attribute;
 }
 #endif
+
 } // namespace
 
 MeshParser::MeshParser()
@@ -187,8 +188,28 @@ bool MeshParser::parse(const QJsonArray &meshArray, GLTF2Context *context)
                 return false;
             }
 
+            const auto primitiveType = static_cast<Qt3DRender::QGeometryRenderer::PrimitiveType>(primitivesObject.value(KEY_MODE).toInt(GL_TRIANGLES));
+            if (primitiveType == Qt3DRender::QGeometryRenderer::Triangles ||
+                primitiveType == Qt3DRender::QGeometryRenderer::TriangleStrip ||
+                primitiveType == Qt3DRender::QGeometryRenderer::TriangleFan) {
+                const auto &attributes = geometry->attributes();
+                const bool hasTangent = std::find_if(std::begin(attributes),
+                                                     std::end(attributes),
+                                                     [](const Qt3DRender::QAttribute *attr) {
+                                                         return attr->name() == Qt3DRender::QAttribute::defaultTangentAttributeName();
+                                                     }) != std::end(attributes);
+
+                if (!hasTangent) {
+                    auto tangentAttr = Kuesa::GLTF2Import::MeshParserUtils::createTangentAttribute(geometry.get(), primitiveType);
+                    if (tangentAttr) {
+                        tangentAttr->setName(Qt3DRender::QAttribute::defaultTangentAttributeName());
+                        geometry->addAttribute(tangentAttr);
+                    }
+                }
+            }
+
             Qt3DRender::QGeometryRenderer *renderer = new Qt3DRender::QGeometryRenderer;
-            renderer->setPrimitiveType(static_cast<Qt3DRender::QGeometryRenderer::PrimitiveType>(primitivesObject.value(KEY_MODE).toInt(GL_TRIANGLES)));
+            renderer->setPrimitiveType(primitiveType);
             renderer->setGeometry(geometry.release());
             primitive.primitiveRenderer = renderer;
             primitive.materialIdx = primitivesObject.value(KEY_MATERIAL).toInt(-1);
--- a/src/core/gltf2importer/meshparser.cpp
+++ b/src/core/gltf2importer/meshparser.cpp
@@ -35,6 +35,8 @@
 #include "gltf2context_p.h"
 #include "kuesa_p.h"
 
+#include "meshparser_utils_p.h"
+
 #include <QJsonObject>
 #include <QJsonArray>
 #include <QDebug>
@@ -160,31 +162,34 @@ bool MeshParser::parse(const QJsonArray &meshArray, GLTF2Context *context)
             const QJsonObject &primitivesObject = primitivesArray[primitiveId].toObject();
 
             bool hasColorAttr = false;
-            Qt3DRender::QGeometry *geometry = new Qt3DRender::QGeometry();
+            auto geometry = std::make_unique<Qt3DRender::QGeometry>();
 
 #if defined(KUESA_DRACO_COMPRESSION)
             const QJsonObject extensions = primitivesObject.value(KEY_EXTENSIONS).toObject();
 
             // Draco Extensions
             if (extensions.contains(KEY_KHR_DRACO_MESH_COMPRESSION_EXTENSION)) {
-                if (!geometryDracoFromJSON(geometry, primitivesObject, hasColorAttr) &&
+                if (!geometryDracoFromJSON(geometry.get(), primitivesObject, hasColorAttr) &&
                     geometry->attributes().isEmpty()) {
-                    delete geometry;
                     return false;
                 }
             } else
 #endif
             {
-                if (!geometryFromJSON(geometry, primitivesObject, hasColorAttr) &&
+                if (!geometryFromJSON(geometry.get(), primitivesObject, hasColorAttr) &&
                     geometry->attributes().isEmpty()) {
-                    delete geometry;
                     return false;
                 }
             }
 
+            if (!Kuesa::GLTF2Import::MeshParserUtils::geometryIsGLTF2Valid(geometry.get())) {
+                qCWarning(kuesa) << QLatin1Literal("Geometry doesn't meet glTF 2.0 requirements");
+                return false;
+            }
+
             Qt3DRender::QGeometryRenderer *renderer = new Qt3DRender::QGeometryRenderer;
             renderer->setPrimitiveType(static_cast<Qt3DRender::QGeometryRenderer::PrimitiveType>(primitivesObject.value(KEY_MODE).toInt(GL_TRIANGLES)));
-            renderer->setGeometry(geometry);
+            renderer->setGeometry(geometry.release());
             primitive.primitiveRenderer = renderer;
             primitive.materialIdx = primitivesObject.value(KEY_MATERIAL).toInt(-1);
             primitive.hasColorAttr = hasColorAttr;
--- a/src/core/gltf2importer/meshparser_utils.cpp
+++ b/src/core/gltf2importer/meshparser_utils.cpp
@@ -113,6 +113,241 @@ int vertexBaseTypeSize(Qt3DRender::QAttribute::VertexBaseType vertexBaseType)
         Q_UNREACHABLE();
     }
 }
+
+struct Attribute {
+    QByteArray bufferData;
+    unsigned int byteOffset;
+    unsigned int byteStride;
+};
+
+// Chances to visit the same face again are high, so cache the state of the last search
+struct FindVertexIndicesInFaceHelper {
+
+    FindVertexIndicesInFaceHelper()
+        : lastFace(-1)
+    {
+    }
+
+    const std::array<unsigned int, 3> &operator()(int iFace) const
+    {
+        if (iFace == lastFace)
+            return verticesForLastFace;
+
+        lastFace = iFace;
+        const auto data = indexAttribute.bufferData;
+        const auto iFaceHead = data.mid(indexAttribute.byteOffset + iFace * indexAttribute.byteStride);
+        switch (vertexBaseType) {
+        case Qt3DRender::QAttribute::UnsignedByte: {
+            const auto *typedIndices = reinterpret_cast<const unsigned char *>(iFaceHead.data());
+            verticesForLastFace = { typedIndices[0], typedIndices[1], typedIndices[2] };
+            break;
+        }
+        case Qt3DRender::QAttribute::UnsignedShort: {
+            const auto *typedIndices = reinterpret_cast<const unsigned short *>(iFaceHead.data());
+            verticesForLastFace = { typedIndices[0], typedIndices[1], typedIndices[2] };
+            break;
+        }
+        case Qt3DRender::QAttribute::UnsignedInt: {
+            const auto *typedIndices = reinterpret_cast<const unsigned int *>(iFaceHead.data());
+            verticesForLastFace = { typedIndices[0], typedIndices[1], typedIndices[2] };
+            break;
+        }
+        default:
+            Q_UNREACHABLE();
+        }
+        return verticesForLastFace;
+    }
+
+    Attribute indexAttribute;
+    Qt3DRender::QAttribute::VertexBaseType vertexBaseType;
+    mutable int lastFace;
+    mutable std::array<unsigned int, 3> verticesForLastFace;
+};
+
+struct MikkTSpaceUserData {
+
+    Attribute indexAttribute;
+    Attribute positionAttribute;
+    Attribute uvAttribute;
+    Attribute normalAttribute;
+
+    int nFaces;
+    int nVertices;
+    Qt3DRender::QGeometryRenderer::PrimitiveType primitiveType;
+    Qt3DRender::QAttribute::VertexBaseType vertexBaseTypeForUVAttribute;
+
+    FindVertexIndicesInFaceHelper vertexIndicesFinder;
+    QByteArray tangentBufferData;
+};
+
+MikkTSpaceUserData precomputeMikkTSpaceUserData(Qt3DRender::QGeometry *geometry, Qt3DRender::QGeometryRenderer::PrimitiveType primitiveType)
+{
+    // Precompute some data
+    MikkTSpaceUserData userData;
+
+    const auto &attributes = geometry->attributes();
+    for (const auto attr : attributes) {
+        if (attr->attributeType() == Qt3DRender::QAttribute::IndexAttribute) {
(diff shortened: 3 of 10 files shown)


src/core/gltf2importer/gltf2importer.pri
```diff
@@ -43,7 +43,8 @@ SOURCES += \
     $$PWD/animationparser.cpp \
     $$PWD/sceneparser.cpp \
     $$PWD/materialparser.cpp \
-    $$PWD/skinparser.cpp
+    $$PWD/skinparser.cpp \
+    $$PWD/mikktspace/mikktspace.c
```
er-vin: Could it go in a 3rdparty folder inside of src?

src/core/gltf2importer/mikktspace/mikktspace.c
```diff
@@ -0,0 +1,1843 @@
+/** \file mikktspace/mikktspace.c
+ *  \ingroup mikktspace
+ */
+/**
+ *  Copyright (C) 2011 by Morten S. Mikkelsen
+ *
+ *  This software is provided 'as-is', without any express or implied
```
er-vin: I have some concerns about the license here. Since I'm not a lawyer, I've a hard time to figure out if it's really compatible with AGPLv3 in practice. My gut feeling is that it is... but my guts aren't a recognized lawyer.

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+        return 1;
+    case Qt3DRender::QAttribute::UnsignedShort:
+        return 2;
+    case Qt3DRender::QAttribute::UnsignedInt:
+        return 4;
+    }
+}
+
+
+std::array<unsigned int, 3> findVertexIndicesInFace(const Qt3DRender::QAttribute *indexAttr, const int iFace) {
```
er-vin: Did you run clang-format? This curly brace should be on its own line.

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+    const auto byteOffset = indexAttr->byteOffset();
+    const auto byteStride = indexAttr->byteStride() == 0 ?
+                            3*vertexBaseTypeSize(indexAttr->vertexBaseType()) :
```
er-vin: Spaces around *

There are more of those, but I'll ignore style issues in the rest of this review. Really, get the clang-format integration with git commit or your IDE so that it does it for you.

src/core/gltf2importer/meshparser.cpp
er-vin: Also, dare I ask for unit tests? :-)

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+        const auto *userData = reinterpret_cast<MikkTSpaceUserData*>(pContext->m_pUserData);
+        return userData->indexAttribute->count() / 3;
+    };
+
+    interface.m_getNumVerticesOfFace = [](const SMikkTSpaceContext *pContext, const int iFace) ->int {
+        return 3;
+    };
+
+    interface.m_getPosition = [] (const SMikkTSpaceContext *pContext,
+                    float fvPosOut[],
```
er-vin: Indentation looks wrong here

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+        fvPosOut[0] = typedPositionHead[0];
+        fvPosOut[1] = typedPositionHead[1];
+        fvPosOut[2] = typedPositionHead[2];
+    };
+
+    interface.m_getTexCoord = [] (const SMikkTSpaceContext *pContext,
+                    float fvPosOut[],
+                    const int iFace,
+                    const int iVertex) {
+        const auto *userData = reinterpret_cast<MikkTSpaceUserData*>(pContext->m_pUserData);
```
er-vin: Lots of duplication around those m_get*, also lots of the work is done more than once. I think we should aim at de-duplicating and reifying the computation in the factored code to make it faster. It's all based on already known addresses and properties of the various attributes at the creation of those lambda, let's lock the lambda behavior at that creation time.

In most of those the only missing information is vertexIndex, all the rest should be already known.

src/core/gltf2importer/meshparser.cpp
jcelerier: could be the array declaration in the lambda parameters that threw the auto-indenter out ?

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+        const auto vertexIndices = findVertexIndicesInFace(userData->indexAttribute, iFace);
+        const auto vertexIndex = vertexIndices[iVertex];
+        const auto vertexBuffer = userData->positionAttribute->buffer();
+        const auto byteOffset = userData->positionAttribute->byteOffset();
+        const auto byteStride = userData->positionAttribute->byteStride() == 0 ?
+                                3*4 : // Position vertex base type must be float
```
jcelerier: maybe an assert to ensure this ?

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+                                    const float fvTangent[],
+                                    const float fSign,
+                                    const int iFace,
+                                    const int iVertex) {
+        auto *userData = reinterpret_cast<MikkTSpaceUserData*>(pContext->m_pUserData);
+        const auto vertexIndices = findVertexIndicesInFace(userData->indexAttribute, iFace);
+        const auto vertexIndex = vertexIndices[iVertex];
+        auto vertexBuffer = userData->tangentAttribute->buffer();
+        const auto byteOffset = 0;
+        const auto byteStride = 4*4;
```
jcelerier: the 4*4 looks fairly magic :p

src/core/gltf2importer/mikktspace/mikktspace.c
jcelerier: it's the zlib license so I am pretty sure it is fine - it's already used in various Qt 3rdparty dependencies

jjcasmar: Who should we ask? Looks like is possible, but an opinion from an expert would be nice.

src/core/gltf2importer/meshparser.cpp
jjcasmar: Looks like clang-format was disabled

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+
+int vertexBaseTypeSize(Qt3DRender::QAttribute::VertexBaseType vertexBaseType)
+{
+    switch (vertexBaseType) {
+    case Qt3DRender::QAttribute::UnsignedByte:
+        return 1;
+    case Qt3DRender::QAttribute::UnsignedShort:
+        return 2;
+    case Qt3DRender::QAttribute::UnsignedInt:
+        return 4;
```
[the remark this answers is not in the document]
jjcasmar: glTF requires this attribute types to ocupy this space. As you said, there is a problem if in a particular system UnsignedInt is an unsigned int which occupies 2bytes for example. Should we use uint_32t and this kind of types?

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+    const auto buffer = indexAttr->buffer();
+    const auto &data = buffer->data();
+    const auto iFaceHead = data.mid(byteOffset + iFace*byteStride);
+    std::array<unsigned int, 3> vertexIndices;
+    switch (indexAttr->vertexBaseType()) {
+    case Qt3DRender::QAttribute::UnsignedByte: {
+        const auto *typedIndices = reinterpret_cast<const unsigned char *>(iFaceHead.data());
+        vertexIndices[0] = typedIndices[0];
+        vertexIndices[1] = typedIndices[1];
+        vertexIndices[2] = typedIndices[2];
```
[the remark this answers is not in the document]
jjcasmar: The face exists because I have create the buffers. If something is wrong, should be signaled before.

src/core/gltf2importer/meshparser.cpp
jjcasmar: Yes, but not here. Attributes have a specific type required by gltf. We should check there that we are using the correct types.

src/core/gltf2importer/gltf2importer.pri
jjcasmar: Im not sure how to do this. Do I have to add a new .pri file?

src/core/gltf2importer/meshparser.cpp
jjcasmar: So Im adding some helper structs that can be used to cache some searchs, have less duplication and make it faster

src/core/gltf2importer/gltf2importer.pri
er-vin: Probably best... but maybe a bit involved for just a file. Or you could use relative paths too. Might have to do something regarding the include path in any cases.

jjcasmar:
I have redo much of the work in this PR. First of all, I have added some checks to verify that the attributes that we generate are valid according to glTF2.0. This is necessary as Im assuming that the attributes have specific sizes and data types. 

I have precompute most of the data that is needed in the generation of tangents and added a helper structure to avoid finding several times the vertex indices for a particular face.

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +210,284 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+        return verticesForLastFace;
+    }
+
+    Attribute indexAttribute;
+    Qt3DRender::QAttribute::VertexBaseType vertexBaseType;
+    mutable int lastFace;
+    mutable std::array<unsigned int, 3> verticesForLastFace;
+};
+
+struct MikkTSpaceUserData {
```
mkrus: clang reports warnings about adding padding to keep memory alignment to members of this struct. Moving `primitiveType` to before `vertexIndicesFinder` fixes those.

This (along with a lot of the pre-existing code in this file) produces warnings of signed/unsigned conversion in places that mix int and uint. `nFaces` and `nVertices` should be uint here I think

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +210,284 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+
+// Chances to visit the same face again are high, so cache the state of the last search
+struct FindVertexIndicesInFaceHelper {
+
+    FindVertexIndicesInFaceHelper()
+        : lastFace(-1)
+    {
+    }
+
+    const std::array<unsigned int, 3> &operator()(int iFace) const
```
mkrus: #include <array> is missing for me...

src/core/gltf2importer/mikktspace/mikktspace.c
er-vin: Oh right, that's the zlib license. Somehow didn't recognize it. We're fine then.

src/core/gltf2importer/meshparser.cpp
jjcasmar: You are right. I have rename it

src/core/gltf2importer/meshparser.cpp
jjcasmar: I have added a new commit to the PR that checks if the mesh is valid. That way, when we generate the tangents, we have already checked all this.

src/core/gltf2importer/meshparser.cpp
```diff
@@ -123,6 +125,201 @@ Qt3DRender::QAttribute *decodeAttribute(const draco::PointCloud *pointCould,
...
+                            indexAttr->byteStride();
+
+    const auto buffer = indexAttr->buffer();
+    const auto &data = buffer->data();
+    const auto iFaceHead = data.mid(byteOffset + iFace*byteStride);
+    std::array<unsigned int, 3> vertexIndices;
+    switch (indexAttr->vertexBaseType()) {
+    case Qt3DRender::QAttribute::UnsignedByte: {
+        const auto *typedIndices = reinterpret_cast<const unsigned char *>(iFaceHead.data());
+        vertexIndices[0] = typedIndices[0];
```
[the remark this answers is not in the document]
jjcasmar: I dont understand the second comment

src/core/gltf2importer/meshparser.cpp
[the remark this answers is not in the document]
jjcasmar: It looks like Qt doesnt work on machines that have strange sizes for some data types. I have found at least a check for float and int in qt codebase

jjcasmar:
I have moved the validation and tangent creation code to meshparser_utils.cpp

lemirep:
Could you had a reference to the 3rd party in the Readme please

src/core/animationplayer.cpp
```diff
@@ -260,6 +262,9 @@ AnimationPlayer::AnimationPlayer(Qt3DCore::QNode *parent)
     });
 
     connect(m_animator, &QClipAnimator::runningChanged, this, &AnimationPlayer::runningChanged);
+    connect(m_animator, &QClipAnimator::runningChanged, []() {
```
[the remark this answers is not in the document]
jjcasmar: In fact, this shouldn't be here at all. Was some debugging code for other patch

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -0,0 +1,381 @@
...
+        return { Qt3DRender::QAttribute::Float,
+                 Qt3DRender::QAttribute::UnsignedByte,
+                 Qt3DRender::QAttribute::UnsignedShort };
+
+    return { Qt3DRender::QAttribute::UnsignedByte,
+             Qt3DRender::QAttribute::UnsignedShort,
+             Qt3DRender::QAttribute::UnsignedInt };
+}
+
+std::vector<uint> validVertexSizesForAttribute(const QString &attributeName)
```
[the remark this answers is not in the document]
jjcasmar: I would prefer to leave it like this. I would need to take into account that -1 in the check after this, which makes it less clear. In most cases this std::vector is not going to perform any allocation anyway, no?

src/core/gltf2importer/mikktspace/mikktspace.h
```diff
@@ -0,0 +1,145 @@
...
+ *  2. Altered source versions must be plainly marked as such, and must not be
+ *     misrepresented as being the original software.
+ *  3. This notice may not be removed or altered from any source distribution.
+ */
+
+#ifndef __MIKKTSPACE_H__
+#define __MIKKTSPACE_H__
+
+
+#ifdef __cplusplus
```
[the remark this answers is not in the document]
jjcasmar: You are right, we dont need to export it to C. Still, if we modify the file, we need to document it according to the license. I just copy paste this files from the Blender repo.

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -0,0 +1,381 @@
...
+#include <Qt3DRender/QAttribute>
+#include <Qt3DRender/QGeometryRenderer>
+
+#include <array>
+
+#include "mikktspace/mikktspace.h"
+
+namespace {
+
+std::vector<Qt3DRender::QAttribute::VertexBaseType> validVertexBaseTypesForAttribute(const QString &attributeName)
```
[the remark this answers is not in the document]
jjcasmar: Same comment as in the std::array<2> above. Im open to discuss it anyway

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -0,0 +1,381 @@
...
+        }
+        default:
+            Q_UNREACHABLE();
+        }
+        return verticesForLastFace;
+    }
+
+    Attribute indexAttribute;
+    Qt3DRender::QAttribute::VertexBaseType vertexBaseType;
+    mutable int lastFace;
```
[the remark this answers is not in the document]
jjcasmar: Im using GSL recommenation here. We are using a very similar approach in the RangeFinder in Qt3D animation aspect. Maybe I could make them private? 

The function must be const, as Im getting a const ptr to the UserData, which gives me a const view of this object. 

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-fct

src/core/gltf2importer/mikktspace/mikktspace.h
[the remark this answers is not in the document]
lemirep: Alright, let's keep it as is then

src/core/gltf2importer/meshparser_utils.cpp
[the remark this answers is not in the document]
lemirep: Well std::vector is always going to perform an allocation. What it does though is probably preallocate a given capacity so that it doesn't have to reallocate every time you insert a new elements. So I guess in the end the question is more about allocation time vs if branch. Note that QVarLengthArray might be interesting in this case as you can specify the maximum amount it should contain but fill in less elements than that if you want to which I guess would allow to not do allocations and not have to perform checks.

src/core/gltf2importer/meshparser_utils.cpp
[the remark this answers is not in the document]
lemirep: Ok let's keep it like that then

src/core/gltf2importer/meshparser_utils.cpp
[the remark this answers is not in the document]
jjcasmar: You are right about std::vector. I thought it has some small buffer optimization or something like that in common implementations as std::string has.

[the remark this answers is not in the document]
jjcasmar: I have moved to QVarLengthArray

src/core/gltf2importer/gltf2importer.pri
jjcasmar: Done. Thanks Paul.

src/core/gltf2importer/meshparser.cpp
```diff
@@ -182,8 +185,34 @@ bool MeshParser::parse(const QJsonArray &meshArray, GLTF2Context *context)
                 }
             }
 
+            if (!Kuesa::GLTF2Import::MeshParserUtils::geometryIsGLTF2Valid(geometry)) {
+                qCWarning(kuesa) << QLatin1Literal("Geometry doesn't meet glTF 2.0 requirements");
+                delete geometry;
```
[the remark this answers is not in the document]
jjcasmar: Changed to use a std::unique_ptr

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -0,0 +1,381 @@
+#include "meshparser_utils_p.h"
```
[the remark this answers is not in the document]
jjcasmar: Fixed

jjcasmar:
rebased

PR #41, opened 2019-02-21
WIP: Add an embedded version of Draco

jcelerier:
In particular, this makes problematic windows debug_and_release builds
more tractable.

~~For now I did not add the Draco files explicitely in order to make review easier. They should go in `Kuesa\src\3rdparty\draco\Draco\[content of the draco repo]`. The files in src\3rdparty\draco\generated are currently updated manually.~~

TODO:
- [x] Test on linux
- [x] Test on mac
- [x] Test no-draco build (linux)
- [x] Test system-draco build (linux)

Questions: 
- ~~Should the files be copied or maybe added as a submodule instead ? ~~
- To build draco I had to disable PCH on windows (else <wingdi.h> gets thrown in and this causes build problems due to `#define ERROR` somewhere in this header). Is this problematic ? I tried to split it out in its own static library for it not to interfere with PCH support for the main Kuesa lib but did not manage for now.
- How to handle all the draco build warnings ?


mkrus:
it works for me on mac, all 3 options...

Just needs the actual Draco files (sources and license file, maybe the readme)

jcelerier:
should be good I think

PR #38, opened 2019-02-13
Add a command-line tool for asset compression

jcelerier:
The command-line API is fairly simple for now (it only allows to set the overall quantization level) - I think that as @seanharmer mentioned the correct thing to do would be to have a project file to indicate the whole set of compression parameters.

PR #42, opened 2019-02-21
Export: change API in order to get the output filename

jcelerier:
Before it was not possible to know in which file the compressed buffer was written.
This will be useful to perform statistics for instance (e.g. for #38 or later if we also do it in the UI).


src/core/gltf2exporter/gltf2exporter_p.h
```diff
@@ -90,6 +90,19 @@ class KUESA_PRIVATE_EXPORT GLTF2Exporter : public QObject
     Q_PROPERTY(SceneEntity *scene READ scene WRITE setScene NOTIFY sceneChanged)
 
 public:
+    class KUESA_PRIVATE_EXPORT Export
```
lemirep: isn't it implicitely exported already since the parent class is exported

jcelerier: indeed !

PR #43, opened 2019-02-21
Sparse accessor

mkrus:
data copying logic is flawed right now...

### DIFF
--- a/src/core/gltf2importer/animationparser.cpp
+++ b/src/core/gltf2importer/animationparser.cpp
@@ -88,19 +88,19 @@ quint8 accessorDataTypeToBytes(Qt3DRender::QAttribute::VertexBaseType type)
 
 QByteArray rawDataFromAccessor(const Accessor &accessor, GLTF2Context *ctx)
 {
-    const BufferView &bufferViewData = ctx->bufferView(accessor.bufferViewIndex);
-
-    if (bufferViewData.bufferData.isEmpty()) {
+    if (accessor.bufferData.isEmpty()) {
         qCWarning(kuesa, "No Buffer found for accessor");
         return {};
     }
 
-    const QByteArray bufferData = bufferViewData.bufferData;
+    const BufferView &bufferViewData = ctx->bufferView(accessor.bufferViewIndex);
+
+    const QByteArray bufferData = accessor.bufferData;
 
     // BufferData was generated using the bufferView's byteOffset
     const qint32 byteOffset = accessor.offset;
     const quint8 elemByteSize = accessorDataTypeToBytes(accessor.type);
-    const qint16 byteStride = (bufferViewData.byteStride > 0 ? bufferViewData.byteStride : qint16(accessor.dataSize * elemByteSize));
+    const qint16 byteStride = (!bufferViewData.bufferData.isEmpty() && bufferViewData.byteStride > 0 ? bufferViewData.byteStride : qint16(accessor.dataSize * elemByteSize));
 
     if (byteStride < qint16(elemByteSize * accessor.dataSize)) {
         qCWarning(kuesa, "Buffer Data byteStride doesn't match accessor dataSize and byte size for type");
--- a/src/core/gltf2importer/bufferaccessorparser.cpp
+++ b/src/core/gltf2importer/bufferaccessorparser.cpp
@@ -37,6 +37,7 @@
 #include <QOpenGLContext>
 
 #include "gltf2context_p.h"
+#include "gltf2keys_p.h"
 
 QT_BEGIN_NAMESPACE
 using namespace Kuesa;
@@ -44,15 +45,14 @@ using namespace GLTF2Import;
 
 namespace {
 
-const QLatin1String KEY_BUFFERVIEW = QLatin1Literal("bufferView");
 const QLatin1String KEY_COMPONENTTYPE = QLatin1Literal("componentType");
 const QLatin1String KEY_COUNT = QLatin1Literal("count");
 const QLatin1String KEY_TYPE = QLatin1Literal("type");
 const QLatin1String KEY_NORMALIZED = QLatin1Literal("normalized");
 const QLatin1String KEY_MAX = QLatin1Literal("max");
 const QLatin1String KEY_MIN = QLatin1Literal("min");
-const QLatin1String KEY_BYTEOFFSET = QLatin1Literal("byteOffset");
-const QLatin1String KEY_NAME = QLatin1Literal("name");
+const QLatin1String KEY_SPARSE_INDICES = QLatin1Literal("indices");
+const QLatin1String KEY_SPARSE_VALUES = QLatin1Literal("values");
 
 QVector<float> jsonArrayToVectorOfFloats(const QJsonArray &values)
 {
@@ -60,7 +60,7 @@ QVector<float> jsonArrayToVectorOfFloats(const QJsonArray &values)
     QVector<float> d(nbComponents);
     std::transform(values.begin(), values.end(), d.begin(),
                    [](const QJsonValue &value) -> float {
-                       return value.toDouble(0);
+                       return static_cast<float>(value.toDouble(0));
                    });
     return d;
 }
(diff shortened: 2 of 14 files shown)


mkrus:
fixed copying logic (thanks Juan!).  Need to add unit tests

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -106,6 +109,74 @@ Qt3DRender::QAttribute::VertexBaseType accessorTypeFromJSON(int componentType)
...
+bool copySparseValues(int count,
+                      const QByteArray &indicesBuffer, int indicesOffset,
+                      const QByteArray &valuesBuffer, int valuesOffset,
+                      QByteArray &targetBuffer, int attributeOffset, int stride,
+                      Qt3DRender::QAttribute::VertexBaseType componentType, int numComponents)
+{
+    switch (componentType) {
+    case Qt3DRender::QAttribute::UnsignedByte:
+    case Qt3DRender::QAttribute::Byte:
+        return copySparseValues<IndexType>(count,
```
jjcasmar: do we really need all this switch again? you already have the type needed in the `typename IndexType`, no?

mkrus: that's not the type of the index, it's the type of the data

mkrus:
added a simple test that data loads

src/core/gltf2importer/bufferaccessorparser.cpp
jjcasmar: But arent you doing a switch to call this function and then doing another switch to call the other copySparseValues overload? My point is that at this point you dont need the switch because you can take directly sizeof(IndexType)

mkrus: you need 2 switches. The outer one is on the type of the index, the inner (in `copySparseValues`) is on the type of the values that get copied.

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -106,6 +110,60 @@ Qt3DRender::QAttribute::VertexBaseType accessorTypeFromJSON(int componentType)
...
+    }
+    return true;
+}
+
+template<class IndexType>
+bool copySparseValues(int count, const QByteArray &indicesBuffer, const QByteArray &valuesBuffer,
+                      QByteArray &targetBuffer, int attributeOffset, int stride,
+                      Qt3DRender::QAttribute::VertexBaseType componentType, int numComponents)
+{
+    switch (componentType) {
```
lemirep: wouldn't it be easier to read if in the switch we only computed the valued size allowing to then only have a single call to copySparseValues?

src/core/gltf2importer/animationparser.cpp
```diff
@@ -88,19 +88,19 @@ int accessorDataTypeToBytes(Qt3DRender::QAttribute::VertexBaseType type)
 
 QByteArray rawDataFromAccessor(const Accessor &accessor, GLTF2Context *ctx)
 {
-    const BufferView &bufferViewData = ctx->bufferView(accessor.bufferViewIndex);
-
-    if (bufferViewData.bufferData.isEmpty()) {
+    if (accessor.bufferData.isEmpty()) {
```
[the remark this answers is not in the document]
lemirep: Ah actually, accessor.bufferData is bufferViewData if we have a bufferViewIndex so that's fine.

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
         accessor.max = jsonArrayToVectorOfFloats(json.value(KEY_MAX).toArray());
         accessor.min = jsonArrayToVectorOfFloats(json.value(KEY_MIN).toArray());
         accessor.name = json.value(KEY_NAME).toString();
+
+        if (json.contains(KEY_SPARSE)) {
+            QJsonObject sparse = json.value(KEY_SPARSE).toObject();
```
lemirep: const QJsonObject

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
...
+            if (!sparse.contains(KEY_COUNT) || !sparse.contains(KEY_SPARSE_INDICES) || !sparse.contains(KEY_SPARSE_VALUES)) {
+                qCWarning(kuesa, "Missing sparse data in accessor");
+                return false;
+            }
+
+            accessor.sparseCount = sparse.value(KEY_COUNT).toInt(0);
+
+            QJsonObject sparseIndices = sparse.value(KEY_SPARSE_INDICES).toObject();
```
lemirep: const QJsonObject

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
...
+            accessor.sparseIndices.type = accessorTypeFromJSON(sparseIndices.value(KEY_COMPONENTTYPE).toInt(GL_UNSIGNED_INT));
+            accessor.sparseIndices.bufferViewIndex = sparseIndices.value(KEY_BUFFERVIEW).toInt(-1);
+            accessor.sparseIndices.offset = sparseIndices.value(KEY_BYTEOFFSET).toInt(0);
+
+            QJsonObject sparseValues = sparse.value(KEY_SPARSE_VALUES).toObject();
```
lemirep: const QJsonObject

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
...
+            accessor.sparseValues.bufferViewIndex = sparseValues.value(KEY_BUFFERVIEW).toInt(-1);
+            accessor.sparseValues.offset = sparseValues.value(KEY_BYTEOFFSET).toInt(0);
+
+            if (accessor.sparseIndices.bufferViewIndex == -1 || accessor.sparseValues.bufferViewIndex == -1) {
+                qCWarning(kuesa, "Missing buffers for sparse indices or values in accessor");
+                return false;
+            }
+
+            QByteArray sparseIndicesData = context->bufferView(accessor.sparseIndices.bufferViewIndex).bufferData;
```
lemirep: const QByteArray

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
...
+            if (sparseIndicesData.isEmpty() || sparseValuesData.isEmpty()) {
+                qCWarning(kuesa, "Empty buffers for sparse indices or values in accessor");
+                return false;
+            }
+
+            int stride = 0;
+            if (accessor.bufferData.isEmpty())
+                accessor.bufferData = QByteArray(accessor.dataSize * accessor.count, 0);
+            else {
+                accessor.bufferData.detach();
```
lemirep: I'm not sure I understand why there's a detach here.

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -135,6 +195,86 @@ bool Kuesa::GLTF2Import::BufferAccessorParser::parse(const QJsonArray &bufferAcc
...
+                const auto &bufferView = context->bufferView(accessor.bufferViewIndex);
+                stride = bufferView.byteStride;
+            }
+
+            switch (accessor.sparseIndices.type) {
+            case Qt3DRender::QAttribute::UnsignedShort:
```
lemirep: shouldn't the Byte/UnsignedByte types also be handled? specs seem to say it's possible unless I've misunderstood

tests/auto/gltfparser/tst_gltfparser.cpp
```diff
@@ -855,6 +855,27 @@ private Q_SLOTS:
         // THEN
         QVERIFY(res != nullptr);
     }
+
+    void checkSparseAccessors()
+    {
```
lemirep: Would it make sense to test against a file that has a sparse accessor that doesn't reference a bufferview to ensure the 0 filled generation works?

src/core/gltf2importer/bufferaccessorparser.cpp
mkrus: good call

src/core/gltf2importer/bufferaccessorparser.cpp
mkrus: we're going to modify the values in the bufferView, but the original content might be used in some other accessor. Sparse values are an optimisation to only store (in the gltf file) the delta between the original and modified content. At runtime you need both states

src/core/gltf2importer/bufferaccessorparser.cpp
mkrus: oops

tests/auto/gltfparser/tst_gltfparser.cpp
mkrus: done

mkrus: done

src/core/gltf2importer/bufferaccessorparser.cpp
```diff
@@ -42,6 +42,7 @@ QT_BEGIN_NAMESPACE
 using namespace Kuesa;
 using namespace GLTF2Import;
 
+
```
jjcasmar: extra line?

mkrus: gone

tests/auto/gltfparser/tst_gltfparser.cpp
```diff
@@ -855,6 +855,45 @@ private Q_SLOTS:
...
+
+        auto accessor = parser.context()->accessor(1);
+        QCOMPARE(accessor.sparseCount, 3);
+        QCOMPARE(accessor.sparseValues.bufferViewIndex, 3);
+        QCOMPARE(accessor.sparseIndices.bufferViewIndex, 2);
+        QVERIFY(!accessor.bufferData.isEmpty());
+
+        QCOMPARE(accessor.bufferViewIndex, bufferView);
+        const float *data = reinterpret_cast<const float*>(accessor.bufferData.constData());
+        QCOMPARE(data[3], originalValue);
```
lemirep: I get a crash when exiting that test function for "Without Original BufferView" (not always seems to happen randomly). It seems there's a double free occurring sometimes when the SceneEntity gets destroyed. I'm not sure how this patch actually affects this.

PR #44, opened 2019-02-22
Exporter: support exporting files with embedded base64 buffers

jcelerier:
Also, I'm wondering if the export UI shouldn't allow the user to choose if he wants to embed existing data maybe ?


mkrus:
can come in another change but I do think it should have a UI to let the user choose what to do (instead of just replicating the existing.
Not sure if you want to let the user decide if images should be inlined or in a buffer. Maybe do what's easiest to implement :)

PR #45, opened 2019-02-25
Refactor URI kind detection

jcelerier:
Also, this fixes the case of gltf buffers referenced by qrc:/buffer.bin urls.


lemirep:
Looks good to me, just waiting to see if CI complains

jcelerier:
hmm, when going through the tests I'm wondering : is Kuesa expected to support cases where some files are stored in a .qrc and some files aren't ? 

Right now it seems that it will only work if all files are in qrc or all files are in the filesystem, so I would like to know if it makes sense to still lookup in the qrc if the gltf contains a `qrc:///blah.png` uri.

lemirep:
Yes you should be able to reference from both a qrc and/or a file on the file system though in that case we would expect the QUrl to be something like "file:///home/..../somefile.png". If the url scheme is qrc:// then we look in the resource, if it's file:/// on the file system. We don't need to try to look for a file on the filesystem if the url scheme is qrc:// though.

jcelerier:
ok, refactored everything a bit more so that both qrc: and file: url schemes in gltf work

src/core/gltf2importer/gltf2uri.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return url;
+}
+
+QString localFileFromGLTFUri(const QString &uri, const QDir &basePath)
+{
+    return absoluteUrlFromGLTFUri(uri, basePath).toLocalFile();
+}
+
+QByteArray dataFromUri(const QString &uri, const QDir &basePath, GLTF2Context *context, bool &success)
```
jjcasmar: Why do we need the context to parse an uri?

mkrus: place to share code... this is used in multiple places

mkrus:
should this include the `data:` URI handling? so can be removed from context?

src/core/gltf2importer/gltf2uri.cpp
jcelerier: hmm, I just checked and the GLTF2Context::parseUri method can be put static so certainly the need for a context instance could be removed

jcelerier: maybe all of this should go in a small Uri namespace ?

jcelerier:
refactored and put in a namespace, I also moved the parseData function out from GLTFContext and put it there.

PR #47, opened 2019-02-26
Add static_asserts to check datatypes have the size we expect them to…

lemirep:
… have

Task-Id: KUE-580

### DIFF
--- a/src/core/gltf2importer/gltf2parser.cpp
+++ b/src/core/gltf2importer/gltf2parser.cpp
@@ -131,6 +131,14 @@ GLTF2Parser::GLTF2Parser(SceneEntity *sceneEntity, bool assignNames)
     , m_defaultSceneIdx(-1)
     , m_assignNames(assignNames)
 {
+    static_assert(sizeof(int) == 4U, "int is expected to be 4 bytes");
+    static_assert(sizeof(uint) == 4U, "unsigned int is expected to be 4 bytes");
+    static_assert(sizeof(short) == 2U, "short is expected to be 2 bytes");
+    static_assert(sizeof(ushort) == 2U, "unsigned short is expected to be 2 bytes");
+    static_assert(sizeof(char) == 1U, "char is expected to be 1 byte");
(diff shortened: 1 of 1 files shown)


src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -131,6 +131,14 @@ GLTF2Parser::GLTF2Parser(SceneEntity *sceneEntity, bool assignNames)
...
     , m_assignNames(assignNames)
 {
+    static_assert(sizeof(int) == 4U, "int is expected to be 4 bytes");
+    static_assert(sizeof(uint) == 4U, "unsigned int is expected to be 4 bytes");
+    static_assert(sizeof(short) == 2U, "short is expected to be 2 bytes");
+    static_assert(sizeof(ushort) == 2U, "unsigned short is expected to be 2 bytes");
+    static_assert(sizeof(char) == 1U, "char is expected to be 1 byte");
+    static_assert(sizeof(uchar) == 1U, "unsigned char is expected to be 1 byte");
+    static_assert(sizeof(float) == 4U, "float is expected to be 4 bytes");
+    static_assert(sizeof(double) == 8U, "double is expected to be 8 bytes");
```
jjcasmar: I think there's no need to check int and double. They are not supported in glTF.

jcelerier:
From the code I saw I'm not sure but I also think that there is the underlying assertion that `char` is signed (and it's not on ARM w/GCC for instance).

e.g. are we sure that the following code : https://github.com/KDAB/kuesa/blob/master/src/core/gltf2importer/animationparser.cpp#L202
and
https://github.com/KDAB/kuesa/blob/master/src/core/gltf2exporter/dracocompressor_p.cpp#L278

will work the same whether char is signed or unsigned ?

jjcasmar:
Should we approve thi

lemirep:
I guess for now it would still work but with probably incorrect values. I'd suggest we create another story to replace the types with the Qt typedef ones (quint8, qint8, quint16, qint16 ...)

PR #48, opened 2019-02-26
default cam for asset viewer

TimoBuskeKdab:
took pauls approach moved to asset pipeline editor directly
and added headroom to the numbers.

worked for the scene i've tested so far...

### DIFF
--- a/tools/assetpipelineeditor/mainwindow.cpp
+++ b/tools/assetpipelineeditor/mainwindow.cpp
@@ -274,8 +274,15 @@ void MainWindow::setCamera(const QString &name)
 
 void MainWindow::viewAll()
 {
-    if (m_camera && m_activeCamera == -1)
-        m_camera->viewAll();
+    if (!m_camera || m_activeCamera != -1)
+        return;
+
+    m_camera->viewAll();
+    QVector3D position = m_camera->position();
+    QVector3D viewCenter = m_camera->viewCenter();
+    float sceneExtent = 2.0f * (viewCenter - position).length();
+    m_camera->setFarPlane(10.0f * sceneExtent); // add some headroom to be able to zoom out
+    m_camera->setNearPlane(0.001f * sceneExtent);
 }
 
 int MainWindow::activeCamera() const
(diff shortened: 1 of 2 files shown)


tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -274,8 +274,15 @@ void MainWindow::setCamera(const QString &name)
...
 {
-    if (m_camera && m_activeCamera == -1)
-        m_camera->viewAll();
+    if (!m_camera || m_activeCamera != -1)
+        return;
+
+    m_camera->viewAll();
+    QVector3D position = m_camera->position();
+    QVector3D viewCenter = m_camera->viewCenter();
+    float sceneExtent = 2.0f * (viewCenter - position).length();
```
jjcasmar: const QVector3D position...
const QVector3D viewCenter...
const float sceneExtent...

mkrus:
viewAll is asynchronous, how does this work?

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -274,8 +274,15 @@ void MainWindow::setCamera(const QString &name)
...
+    m_camera->setFarPlane(10.0f * sceneExtent); // add some headroom to be able to zoom out
+    m_camera->setNearPlane(0.001f * sceneExtent);
 }
```
lemirep: m_camera->viewAll() doesn't do anything immediately (it sends commands to Qt3D but the actual response is received later). I think you're just getting lucky with the default values of position and viewCenter and the factors you've applied. What you could do however is connect a lambda or slot to the signals QCamera::positionChanged and/or viewCenterChanged which should be emitted once the viewAll command has been processed to recompute the sceneExtent.

TimoBuskeKdab: thanks for the info. could be i was lucky. i'll check some more scenes then...

TimoBuskeKdab:
added 3 test scenes to the jira ticket and tested.

PR #49, opened 2019-02-26
Warn the user about the assets that may be overwritten while exporting.

jcelerier:
This saves a list of the files referenced by a .gltf in the gltf context so that we can check them quickly before exporting.

### DIFF
--- a/src/core/gltf2exporter/gltf2exporter_p.cpp
+++ b/src/core/gltf2exporter/gltf2exporter_p.cpp
@@ -664,6 +664,19 @@ void GLTF2Exporter::setContextFromImporter(GLTF2Importer *importer)
     m_context = importer->m_context;
 }
 
+QStringList GLTF2Exporter::overwritableFiles(const QDir &target) const
+{
+    QStringList files;
+    for (const auto& file : m_context->localFiles()) {
+        auto filePath = GLTF2Import::Uri::localFile(file, target);
+        if (!filePath.startsWith(QStringLiteral(":/"))) {
+            if (QFileInfo::exists(filePath))
+                files.append(filePath);
+        }
+    }
+    return files;
+}
+
 auto GLTF2Exporter::saveInFolder(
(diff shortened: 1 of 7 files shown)


src/core/gltf2exporter/gltf2exporter_p.cpp
```diff
@@ -664,6 +664,19 @@ void GLTF2Exporter::setContextFromImporter(GLTF2Importer *importer)
     m_context = importer->m_context;
 }
 
+QStringList GLTF2Exporter::overwritableFiles(const QDir &target) const
+{
+    QStringList files;
+    for (auto file : m_context->localFiles()) {
```
lemirep: won't that cause a container detach since we aren't doing const auto (or const QString &) ?

src/core/gltf2importer/gltf2context_p.h
```diff
@@ -146,6 +146,9 @@ class Q_AUTOTEST_EXPORT GLTF2Context
     const QJsonDocument &json() const;
     void setJson(const QJsonDocument &doc);
 
+    const QVector<QString> &localFiles() const;
```
lemirep: why go for QStringList in overwritableFiles and QVector<QString> here?

jcelerier: good question, hadn't thought about it. which one would be prefferred ? I think that I was influence by many things being QVector<foo> in GLTFContext  but on the other hand, QStringList has nice utilty methods...

src/core/gltf2exporter/gltf2exporter_p.cpp
jcelerier: fixed

src/core/gltf2importer/gltf2context_p.h
jcelerier: went for qstringlist in the latest commit

PR #50, opened 2019-02-27
Fix a potential infinite loop when exporting

jcelerier:
Sorry !


src/core/gltf2exporter/gltf2exporter_p.cpp
```diff
@@ -60,6 +60,7 @@ QString generateUniqueFilename(const QDir &dir, QString filename)
     int num = 1;
     while (dir.exists(filename)) {
         filename = QStringLiteral("%1-%2.%3").arg(basename).arg(num).arg(ext);
+        num++;
```
lemirep: ++num even since you don't need to read back the value immediately :)

jcelerier: done

PR #51, opened 2019-02-27
Use Qt type typedefs

lemirep:
As for instance char might be an unsigned char on Arm.

Task-Id: KUE-580


jjcasmar:
Looks okay to me

PR #55, opened 2019-02-28
Fix unit tests:

lemirep:
- TEXCOORD_1 and TEXCOORD_2 to Qt3DRender::QAttribute name was missing
- Context was missing accessor info for checkForAssetPipelineEditorProperties
- Rewrote geometryIsGLTF2Valid to make it easier to understand

### DIFF
--- a/src/core/gltf2importer/meshparser_utils.cpp
+++ b/src/core/gltf2importer/meshparser_utils.cpp
@@ -348,6 +348,29 @@ SMikkTSpaceInterface createMikkTSpaceInterface()
     interface.m_setTSpace = nullptr;
     return interface;
 }
+
+bool vertexBaseTypeForAttributesAreValid(const QVector<Qt3DRender::QAttribute *> &attributes)
+{
+    for (Qt3DRender::QAttribute *attribute : attributes) {
+        const auto validVertexBaseTypes = validVertexBaseTypesForAttribute(attribute->name());
+        const auto vertexBaseType = attribute->vertexBaseType();
+        if (!validVertexBaseTypes.contains(vertexBaseType))
+            return false;
+    }
+    return true;
+}
+
+bool vertexSizesForAttributesAreValid(const QVector<Qt3DRender::QAttribute *> &attributes)
+{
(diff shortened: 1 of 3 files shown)


jjcasmar:
I would say commit message and changes are unrelated. The change in the is valid function has nothing to do with the tests. Same happens for the TEXCOOR_1/2.

Regarding the refactor of the is valid function, I wouldn't say its easier to understand now. You are defining a lambda, which is invoked directly. Also, you are capturing the attributes and returning from the lambda.  I found the current implementation (based on std::find_if) easy, but if you prefer to modify it, I would suggest going for a raw loop instead of a lambda

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -382,27 +382,25 @@ Qt3DRender::QAttribute *createTangentAttribute(Qt3DRender::QGeometry *geometry,
...
-                                                  [](Qt3DRender::QAttribute *attribute) {
-                                                      const auto validVertexSizes = validVertexSizesForAttribute(attribute->name());
-                                                      const auto vertexSize = attribute->vertexSize();
-                                                      return std::find_if(std::begin(validVertexSizes),
-                                                                          std::end(validVertexSizes),
-                                                                          [vertexSize](uint validVertexSize) {
-                                                                              return vertexSize == validVertexSize;
-                                                                          }) == std::end(validVertexSizes);
-                                                  }) == std::end(attributes);
+    const bool vertexBaseTypesAreValid = [&] () {
```
jjcasmar: I dont find this easier to read. Added a comment in the conversation.

lemirep: What do you want me to do exactly ? The commit is about fixing the unit tests which are currently broken since the introduction of the geometryIsGLTF2Valid leaving the CI in a broken state currently. For the tests to pass, as indicated in the commit message, the first two entries are necessary. I do agree that the last entry in the commit message is more voicing my personal opinion than factuals, I've updated the commit message accordingly and reworked the code following your comments. Please let's try not to hold this patch in review too long

PR #62, opened 2019-03-06
Mac fixes

mkrus:
broke the bundle :(


mkrus:
they are, was rebasing on different branch...

PR #65, opened 2019-03-07
Fix Shader Issue Under ES2 / Windows

TimoBuskeKdab:
got QOpenGLShader::compile(Fragment): ERROR: 0:81: '=' :
assigning non-constant to 'const highp float'

Ticket-Id: KUE-631


TimoBuskeKdab:
> What a crappy compiler but I guess it's within the spec technically since it calls the sqrt function. We could work out the actual value by hand and leave a comment as to what was used to calculate it. That way we can keep this as a constant and avoid a sqrt() operation per fragment.

right. sounds good. i'll change that.

PR #67, opened 2019-03-11
Static analysis fixes

mkrus:


src/core/gltf2exporter/gltf2exporter_p.cpp
```diff
@@ -135,8 +135,8 @@ class GLTF2DracoCompressor
         , m_accessors(rootObject[GLTF2Import::KEY_ACCESSORS].toArray())
         , m_meshes(rootObject[GLTF2Import::KEY_MESHES].toArray())
         , m_images(rootObject[GLTF2Import::KEY_IMAGES].toArray())
-        , m_compressedBufferIndex(m_buffers.size()) // Indice of the added buffer
```
jcelerier: Today I Learned :-)

tools/assetpipelineeditor/meshinspector.cpp
```diff
@@ -105,7 +105,8 @@ void MeshInspector::setMesh(Qt3DRender::QGeometryRenderer *mesh)
     m_primitiveType = m_mesh ? nameForPrimitiveType(m_mesh->primitiveType()) : QString();
 
     if (m_mesh) {
-        for (auto attribute : m_mesh->geometry()->attributes()) {
+        const auto attributes = m_mesh->geometry()->attributes();
```
jcelerier: const auto& ?

tools/assetprocessor/main.cpp
```diff
@@ -137,7 +137,8 @@ int main(int argc, char *argv[])
 
     if (!new_asset.success()) {
         QTextStream errors;
-        for (const auto &err : exporter.errors())
+        const auto reporterErrors = exporter.errors();
```
jcelerier: const auto& ?

src/core/framegraphes/forwardrenderer.cpp
```diff
@@ -339,7 +339,7 @@ ForwardRenderer::ForwardRenderer(Qt3DCore::QNode *parent)
 ForwardRenderer::~ForwardRenderer()
 {
     // unparent the effect subtrees or they'll be deleted twice
-    for (auto framegraph : qAsConst(m_effectFGSubtrees))
+    for (auto &framegraph : qAsConst(m_effectFGSubtrees))
```
jcelerier: const auto & ?

jcelerier: (though if it's only a pointer I wonder what is being gained)

mkrus: it's a `QSharedPointer` so we do save on a ref/unref. Can't be const since `QNode::setParent` is not const

jcelerier: ah, indeed

PR #69, opened 2019-03-11
Use layer.enabled to cache high-triangle buttons

wheeland:
Don't do it for every button, but for groups of controls, to not
incur a texture bind + draw call for every control element.


lemirep (commented):
This will help if the user doesn't touch the menu but if the user start playing with the sliders or buttons this forces a render to texture to happen which then is more costly than we we have right now. Wouldn't it be better to modify StyledSlider/StyledSwitch to use pre-generated images/border images ?

wheeland:
Yes, but we'd have to drop the color animation, or add a custom SG node? The latter seems like a bit too much hassle for a demo?

PR #71, opened 2019-03-11
Introduce music box demo

mpersano:


jcelerier:
Is the song defined in `musicRows` random or does it correspond to something existing ? (so that I can check that the sound generation is correct - e.g. if it's chromatic or diationic)

jcelerier:
when zooming in the skybox disappears for some reason, but maybe this is a more general Kuesa bug ?

jcelerier:
sample player added : https://github.com/KDAB/kuesa/tree/music-box-demo (maybe the PR should be closed and reopened with the local branch ? it will be easier if we are multiple persons to contribute to it).

- the sounds aren't the correct ones yet
- lot of stuff hardcoded (for a demo I think that it is fine but YMMV)
- note -> sample mapping incorrect for now, if it is going to change from song to song it should be somehow added in the qml file and sent to the sampler class

PR #73, opened 2019-03-12
AnimationParser: add additional check

lemirep:
To make sure we have the right number of components for a given channel type

### DIFF
--- a/src/core/gltf2importer/animationparser.cpp
+++ b/src/core/gltf2importer/animationparser.cpp
@@ -173,6 +173,21 @@ QString channelPathToProperty(const QString &path)
     return {};
 }
 
+quint8 expectedComponentsCountForChannel(const QString &channelName)
+{
+    if (channelName == QStringLiteral("translation"))
+        return 3;
+    if (channelName == QStringLiteral("rotation"))
+        return 4;
+    if (channelName == QStringLiteral("scale"))
+        return 3;
+    if (channelName == QStringLiteral("weights"))
+        return 1;
+
+    qCWarning(kuesa) << "Unrecognized animation channel" << channelName;
+    return 0;
+}
+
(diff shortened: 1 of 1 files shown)


src/core/gltf2importer/animationparser.cpp
```diff
@@ -173,6 +173,19 @@ QString channelPathToProperty(const QString &path)
...
+{
+    if (channelName == QStringLiteral("translation"))
+        return 3;
+    if (channelName == QStringLiteral("rotation"))
+        return 4;
+    if (channelName == QStringLiteral("scale"))
+        return 3;
+    if (channelName == QStringLiteral("weights"))
+        return 1;
+    return 0;
```
mkrus: might be good, if the result is 0, to print a warning that the channel name is invalid

src/core/gltf2importer/animationparser.cpp
```diff
@@ -249,6 +262,14 @@ Qt3DAnimation::QChannel AnimationParser::animationChannelDataFromBuffer(const QS
     std::vector<float> valueStamps;
     std::tie(nbComponents, valueStamps) = animationChannelDataFromData(sampler);
 
+    // Check nbComponent and type are what is expected for a given channel
+    const quint8 expectedComponents = expectedComponentsCountForChannel(channelName);
+
```
mkrus: and if warning is printed above, have an early return on 0 result (not printing anything else)

src/core/gltf2importer/animationparser.cpp
lemirep: done

src/core/gltf2importer/animationparser.cpp
lemirep: done

PR #76, opened 2019-03-12
Implement saving and loading camera settings

seanharmer:
Allows to easily get reproducable cameras between runs to be able
to compare images.

### DIFF
--- a/tools/assetpipelineeditor/camerawidget.cpp
+++ b/tools/assetpipelineeditor/camerawidget.cpp
@@ -30,6 +30,33 @@
 #include "ui_camerawidget.h"
 #include "ui_vectorwidget.h"
 
+#include <QFileDialog>
+#include <QFileInfo>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QSettings>
+
+namespace {
+    const QLatin1String LASTPATHSETTING("mainwindow/lastPath");
+
+    QJsonObject vector3DToJson(const QVector3D &v)
+    {
+        QJsonObject json;
+        json[QLatin1String("x")] = static_cast<double>(v.x());
+        json[QLatin1String("y")] = static_cast<double>(v.y());
+        json[QLatin1String("z")] = static_cast<double>(v.z());
+        return json;
+    }
+
+    QVector3D jsonToVector3D(const QJsonObject &json)
+    {
+        QVector3D v(static_cast<float>(json[QLatin1String("x")].toDouble()),
+                    static_cast<float>(json[QLatin1String("y")].toDouble()),
+                    static_cast<float>(json[QLatin1String("z")].toDouble()));
+        return v;
+    }
+}
+
 VectorWidget::VectorWidget(QWidget *parent)
     : QWidget(parent)
     , m_ui(new Ui::VectorWidget)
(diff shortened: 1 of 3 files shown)


tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -30,6 +30,33 @@
...
+#include <QJsonObject>
+#include <QSettings>
+
+namespace {
+    const QLatin1String LASTPATHSETTING("mainwindow/lastPath");
+
+    QJsonObject toJson(const QVector3D &v)
+    {
+        QJsonObject json;
+        json["x"] = static_cast<double>(v.x());
```
lemirep: is "x" a QString or QLatin1String here ?

tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -227,6 +256,61 @@ void CameraWidget::setFoV(double fov)
...
+    settings.setValue(LASTPATHSETTING, fi.absolutePath());
+
+    QFile saveFile(filename);
+    if (!saveFile.open(QIODevice::WriteOnly)) {
+        qWarning("Couldn't open camera settings save file.");
+        return;
+    }
+
+    QJsonObject cameraObject;
+    write(cameraObject);
```
lemirep: write/read are a bit confusing as they are usually related to either the system calls or QIODevice read/write. How about doing QJsonObject cameraObject = toJson(); and fromJson() to restore?

tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -227,6 +256,61 @@ void CameraWidget::setFoV(double fov)
...
+    QFileInfo fi(filename);
+    settings.setValue(LASTPATHSETTING, fi.absolutePath());
+
+    QFile loadFile(filename);
+    if (!loadFile.open(QIODevice::ReadOnly)) {
+        qWarning("Couldn't open camera settings load file.");
+        return;
+    }
+
+    QByteArray savedData = loadFile.readAll();
```
lemirep: const QByteArray

tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -227,6 +256,61 @@ void CameraWidget::setFoV(double fov)
...
+    QJsonDocument loadDoc(QJsonDocument::fromJson(savedData));
```
lemirep: const QJsonDocument

tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -30,6 +30,33 @@
...
+#include <QFileDialog>
+#include <QFileInfo>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QSettings>
+
+namespace {
+    const QLatin1String LASTPATHSETTING("mainwindow/lastPath");
+
+    QJsonObject toJson(const QVector3D &v)
```
lemirep: vector3DToJson() ? or you need to have a templated T fromJson(const QJsonObject) below

tools/assetpipelineeditor/camerawidget.cpp
```diff
@@ -257,3 +341,55 @@ void CameraWidget::updateWidgetValues()
...
+    json["viewCenter"] = toJson(m_camera->viewCenter());
+    json["upVector"] = toJson(m_camera->upVector());
+    json["nearPlane"] = static_cast<double>(m_camera->nearPlane());
+    json["farPlane"] = static_cast<double>(m_camera->farPlane());
+
+    const bool isPerspective = m_camera->projectionType() == Qt3DRender::QCameraLens::PerspectiveProjection;
+    if (isPerspective) {
+        json["projectionType"] = QLatin1String("Perspective");
+        json["fieldOfView"] = static_cast<double>(m_camera->fieldOfView());
+        json["aspectRatio"] = static_cast<double>(m_camera->aspectRatio());
```
TimoBuskeKdab: is this something we want to save and we can load properly?
it doesn't really hurt but may lead to distorted output after loading in a window with different aspect until aspect changes...

seanharmer: Hmm yeah. Let's get it in and see how it behaves - perhaps we can add a follow up to trigger a new fov and aspect ratio or left/right/top/bottom calc to be done somehow.

PR #78, opened 2019-03-13
gles linter

TimoBuskeKdab:
python script for checking gles compatibility of asset files
.dds .png .jpg

todo:
.gltf (mesh size)

required: ImageMagick


lemirep:
The 3rd party linter code might not be compatible at all with python 3 though

TimoBuskeKdab:
Okay, no problem. I just thought it's a good time to switch according to
https://pythonclock.org/
Actually the 3rd party stuff gave me errors using Python 2. But that should be fixable somehow I think.

TimoBuskeKdab:
okay should work now on python 2 + 3

mkrus:
needs some documentation, README.md with info on usage, and details of 3rd party code used...

PR #79, opened 2019-03-14
Morphtargets

lemirep:
- Mesh and Animation GLTF2 parsers updated to retrieve MorphTarget animation
- Actual animation and generation of compatible shader still needs to be done

### DIFF
--- a/src/core/gltf2importer/animationparser.cpp
+++ b/src/core/gltf2importer/animationparser.cpp
@@ -159,6 +159,8 @@ QString channelPathToName(const QString &path)
         return QStringLiteral("Rotation");
     if (path == QStringLiteral("scale"))
         return QStringLiteral("Scale3D");
+    if (path == QStringLiteral("weights"))
+        return QStringLiteral("MorphWeights");
     return {};
 }
 
@@ -170,6 +172,8 @@ QString channelPathToProperty(const QString &path)
         return QStringLiteral("rotation");
     if (path == QStringLiteral("scale"))
         return QStringLiteral("scale3D");
+    if (path == QStringLiteral("weights"))
+        return QStringLiteral("morphWeights");
     return {};
 }
 
@@ -235,7 +239,8 @@ std::tuple<int, std::vector<float>> AnimationParser::animationChannelDataFromDat
 }
 
 Qt3DAnimation::QChannel AnimationParser::animationChannelDataFromBuffer(const QString &channelName,
-                                                                        const AnimationParser::AnimationSampler &sampler) const
+                                                                        const AnimationParser::AnimationSampler &sampler,
+                                                                        const TreeNode &targetNode) const
 {
     auto channel = Qt3DAnimation::QChannel(channelPathToName(channelName));
     const Accessor inputAccessor = m_context->accessor(sampler.inputAccessor);
@@ -276,6 +281,23 @@ Qt3DAnimation::QChannel AnimationParser::animationChannelDataFromBuffer(const QS
 
     if (nbComponents == 0 || valueStamps.empty())
         return channel;
+
+    // When dealing with morph targets we have :
+    // - morphTargetCount * timeStamps.size() keyframe values
+    // - animationChannelDataFromData will set nbComponents to 1 because we are
+    // dealing with a scalar weight type
+    // - the channel should actually expose one scalar weight for each morph
+    // targets
+    // This means we should set nbComponent to be morphTargetCount
+    const bool isMorphTargetWeightChannel = (channelName == QStringLiteral("weights"));
+    if (isMorphTargetWeightChannel) {
+        const Mesh morphTargetMesh = m_context->mesh(targetNode.meshIdx);
+        if (targetNode.meshIdx < 0 || morphTargetMesh.morphTargetCount == 0) {
+            qWarning(kuesa) << "Invalid Mesh for MorphTarget animation";
+            return channel;
+        }
+        nbComponents = morphTargetMesh.morphTargetCount;
+    }
 
     Qt3DAnimation::QKeyFrame::InterpolationType interpolationType = Qt3DAnimation::QKeyFrame::LinearInterpolation;
 
@@ -455,11 +477,12 @@ AnimationParser::channelFromJson(const QJsonObject &channelObject) const
     const QJsonValue samplerValue = channelObject.value(KEY_SAMPLER);
     const QJsonObject targetObject = channelObject.value(KEY_TARGET).toObject();
     const QString path = targetObject.value(KEY_PATH).toString();
-    const int targetNode = targetObject.value(KEY_NODE).toInt();
+    const int targetNodeIdx = targetObject.value(KEY_NODE).toInt(-1);
     const AnimationSampler sampler = m_samplers.at(samplerValue.toInt());
+    const TreeNode targetNode = m_context->treeNode(targetNodeIdx);
 
-    channel = animationChannelDataFromBuffer(path, sampler);
-    channel.setName(channel.name() + QStringLiteral("_") + QString::number(targetNode));
+    channel = animationChannelDataFromBuffer(path, sampler, targetNode);
+    channel.setName(channel.name() + QStringLiteral("_") + QString::number(targetNodeIdx));
 
     if (channel.channelComponentCount() == 0) {
         qCWarning(kuesa, "Channel doesn't have components");
@@ -474,11 +497,6 @@ AnimationParser::mappingFromJson(const QJsonObject &channelObject) const
 {
     const QJsonObject targetObject = channelObject.value(KEY_TARGET).toObject();
     const QString path = targetObject.value(KEY_PATH).toString();
-
-    // TODO If we find a weight mapper, return a default constructed mapping
-    // Qt3D doesn't support morph targets yet
-    if (path == QStringLiteral("weights"))
-        return std::make_tuple(true, ChannelMapping());
 
     const QJsonValue nodeValue = targetObject.value(KEY_NODE);
     ChannelMapping mapping;
@@ -582,13 +600,6 @@ bool AnimationParser::parse(const QJsonArray &animationsArray, GLTF2Context *con
                 qCWarning(kuesa, "An animation mapping is incorrect");
                 return false;
             }
-
-            // TODO Weights mappings are default constructed.
-            // Detect them with the targetNodeId default value and discard them so we don't try to play them
-            // Qt3D doesn't support morph targets yet
-            if (mapping.targetNodeId == -1)
-                continue;
-
             animation.mappings.push_back(mapping);
         }
 
--- a/src/core/gltf2importer/animationparser.cpp
+++ b/src/core/gltf2importer/animationparser.cpp
@@ -93,7 +93,7 @@ QByteArray rawDataFromAccessor(const Accessor &accessor, GLTF2Context *ctx)
         return {};
     }
 
-    const BufferView &bufferViewData = ctx->bufferView(accessor.bufferViewIndex);
+    const BufferView bufferViewData = ctx->bufferView(accessor.bufferViewIndex);
 
     const QByteArray bufferData = accessor.bufferData;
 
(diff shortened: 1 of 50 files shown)


jalbamon (commented):
Documentation looks in order

src/core/gltf2importer/animationparser.cpp
```diff
@@ -277,6 +282,22 @@ Qt3DAnimation::QChannel AnimationParser::animationChannelDataFromBuffer(const QS
     if (nbComponents == 0 || valueStamps.empty())
         return channel;
 
+    // When dealing with morph targets we should have
+    // nbComponents * morphTargetCount * timeStamps.size() keyframe values
+    // nbComponents is 1 because we are dealing with a scalar weight type
+    // but the channel should actually expose one scalar weight for each morph targets
+    // -> one channel component for each morph target in the mesh
+    // 1 - n possibly
+    const bool isMorphTargetWeightChannel = (channelName == QStringLiteral("weights"));
```
jalbamon: I'm a little confused by the last 3 lines of the comment, but that could be because I have no idea about morph targets.  Just check to make sure it makes sense.

lemirep: Right, I found it confusing it as well. Reworded it.

lemirep:
Updated with latest dev. Substitution for the if node had to be fixed for the shader generation to work correctly and not break multi UV support. Tested along with https://codereview.qt-project.org/#/c/258754/

src/core/morphcontroller.cpp
```diff
@@ -0,0 +1,393 @@
...
+        &MorphController::weight5,
+        &MorphController::weight6,
+        &MorphController::weight7
+    };
+
+    QVariantList weights;
+    weights.reserve(m_count);
+
+    for (int i = 0; i < m_count; ++i)
+        weights.push_back((this->*weightGetterFuncs[i])());
```
DaveeFTW: This seems a little complicated. Would it not be simpler to just copy from `m_weights` instead?

src/core/morphcontroller.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+    return weights;
+}
+
+void MorphController::setWeight0(float weight0)
+{
+    if (weight0 == m_weights[0])
```
DaveeFTW: Use a fuzzy comparison here

src/core/gltf2importer/meshparser.cpp
```diff
@@ -219,6 +247,74 @@ bool MeshParser::parse(const QJsonArray &meshArray, GLTF2Context *context)
             primitive.hasColorAttr = hasColorAttr;
         }
 
+        // All primitives in a mesh are required to declare the same number of
```
jjcasmar: Is this glTF requirement or Kuesa?

src/core/metallicroughnessmaterial.cpp
```diff
@@ -322,6 +323,15 @@ WrappedSignal<OutputType> wrapParameterSignal(MetallicRoughnessMaterial *self, S
     \since 1.1
  */
 
+/*!
+    \property morphController
+
+    The morph controller instance used to control morph targets weights.
+    This can be null if a material is not used without morph targets.
```
jjcasmar: "a material is not used without morph targets" means a material that uses morph targets?

src/core/shaders/es2/kuesa_morphtargets_normal.inc.vert
```diff
@@ -0,0 +1,46 @@
...
+attribute vec3 vertexNormal_2;
+attribute vec3 vertexNormal_3;
+attribute vec3 vertexNormal_4;
+
+// Note: it is assumed the morphWeights structure uniform has been defined
+// elsewhere and included prior to this
+
+vec3 kuesa_morphNormal(const in vec3 vNormal)
+{
+    vec3 norm = vNormal;
```
jjcasmar: if the primitive only contains 2 morphtargets, isnt this doing unneded computations?

lemirep: yes but handling it this way is I think for now the best compromise between performance/maintenance until we find a way to provide multiple shader node tree description to the shader builder in a sane way.

src/core/metallicroughnessmaterial.cpp
lemirep: Fixed, should be "with" and not "without".

src/core/gltf2importer/meshparser.cpp
lemirep: I think it's an implicit requirement from glTF (Spec says: All primitives are required to list the same number of targets in the same order)

src/core/morphcontroller.cpp
lemirep: I don't think we want to do that, we want to compare the exact values rather than whether the two are close values.

src/core/morphcontroller.cpp
lemirep: Good point

src/core/gltf2importer/animationparser.cpp
```diff
@@ -361,8 +383,8 @@ Qt3DAnimation::QChannel AnimationParser::animationChannelDataFromBuffer(const QS
 
 bool AnimationParser::checkSamplerJSON(const QJsonObject &samplerObject) const
 {
-    const QJsonValue &inputValue = samplerObject[KEY_INPUT];
-    const QJsonValue &outputValue = samplerObject[KEY_OUTPUT];
+    const QJsonValue inputValue = samplerObject.value(KEY_INPUT);
```
jjcasmar: probably for future patches, we should avoid modifiying this kind of stuff. It add noise to the feature we are trying to review

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -864,14 +868,50 @@ void GLTF2Parser::generateTreeNodeContent()
...
+                // Create a Morph Controller if we have morph targets
+                {
+                    if (hasMorphTargets) {
+                        // Initiliaze the Morph Controller with the default weights
+                        const QVector<float> nodeDefaultWeight = node.morphTargetWeights;
+                        const QVector<float> meshDefaultWeights = meshData.morphTargetWeights;
+                        const quint8 morphTargetCount = std::min(meshData.morphTargetCount, quint8(8));
+
+                        if (meshData.morphTargetCount > 8)
+                            qCWarning(kuesa) << "Kuesa only supports up to 8 morph targets per mesh";
```
jjcasmar: does it make sense to continue if we are not going to animate all the morph targets?

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -75,13 +86,28 @@ QVarLengthArray<Qt3DRender::QAttribute::VertexBaseType, 3> validVertexBaseTypesF
...
+    }
+
     // Standard Attributes
     if (attributeName == Qt3DRender::QAttribute::defaultPositionAttributeName())
         return { 3 };
     if (attributeName == Qt3DRender::QAttribute::defaultNormalAttributeName())
         return { 3 };
     if (attributeName == Qt3DRender::QAttribute::defaultTangentAttributeName())
-        return { 4 };
+        return { 3, 4 };
```
jjcasmar: why can the tangent have 3 components?

src/core/metallicroughnessmaterial.cpp
```diff
@@ -971,6 +999,14 @@ void MetallicRoughnessMaterial::setToneMappingAlgorithm(MetallicRoughnessEffect:
     m_effect->setToneMappingAlgorithm(algorithm);
 }
 
+void MetallicRoughnessMaterial::setMorphController(MorphController *morphController)
+{
+    if (morphController == m_morphControllerParameter->value().value<MorphController *>())
+        return;
+
+    m_morphControllerParameter->setValue(QVariant::fromValue(morphController));
```
jjcasmar: Im not sure I get this. You are setting a pointer to a custom class for the parameter, which is the uniform in the shader?

jjcasmar:
What happens if we have a mesh with a position morph target and without normals or tangents. In this case we need to generate the normals or tangents for the primitive, but how do we also need to generate a morph target for normals/tangents?

src/core/gltf2importer/meshparser_utils.cpp
lemirep: well because the the morph target sample files are only providing 3

src/core/gltf2importer/gltf2parser.cpp
lemirep: I think so, at least you'll be able to have something working and the warning, rather than the warning and nothing

src/core/metallicroughnessmaterial.cpp
lemirep: MorphController is a ShaderData, setting a ShaderData on a QParameter is possible and is used when you need to set a struct for a uniform

src/core/gltf2importer/animationparser.cpp
lemirep: well it's done as a dedicated commit https://github.com/KDAB/kuesa/pull/79/commits/ae8c202138e5412c6524dec2267c881bd44fc494
it's just that if you review with the all files changes, you don't see that.

src/core/gltf2importer/meshparser_utils.cpp
jjcasmar: Okay, so I guess is just setting the values of the tangents and the base configuration (the fourth component) is unchanged.

jjcasmar:
I have added an issue in the gltf2 specification github to try to answer the normal/tangent generation. Meanwhile, we can go as it is

src/core/shaders/graphs/metallicroughness.frag.json
```diff
@@ -2117,15 +2117,15 @@
                         "major": 2,
                         "minor": 0
                     },
-                    "substitution": "highp $type $output; if ($condition) discard; $output = $value;"
+                    "substitution": "highp $type $output = $value; if ($condition) discard;"
```
jjcasmar: With your last changes on the shader generator, isnt it more efficient to leave the assignment (which is going to perform the computation) after checking if we need to discard?

lemirep: afaik I had to change as the shader generator expect something like * var = assignment; *
So had to change the expression to make it work though I suspect this might result in the discard part completely being removed. Which would explain why the opacity is not working properly recently

lemirep: Fixed the discard if. Any substituion that requires more than one line or statement (more than one ;) should call a function. The qshadergenerator might not read it correctly or optimize it away when it shouldn't.

src/core/shaders/es2/kuesa_morphtargets_normal.inc.vert
```diff
@@ -0,0 +1,42 @@
...
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+attribute vec3 vertexNormal_1;
+attribute vec3 vertexNormal_2;
+
+// Note: it is assumed the morphWeights structure uniform has been defined
+// elsewhere and included prior to this
+
+vec3 kuesa_morphNormal(const in vec3 vNormal)
```
mkrus: doesn't this need precision qualifiers?  (same with other shaders)

src/core/shaders/graphs/metallicroughness.frag.json
jjcasmar: nice. probably should be documented on the node editor and shader builder in qtbase.

src/core/shaders/es2/kuesa_morphtargets_normal.inc.vert
lemirep: shouldn't be needed, kuesa_morphtargets_weights.inc.frag defines precision highp float;
precision highp vec3; and is always supposed to be included before the other morphtarget includes

mkrus (approved):

PR #80, opened 2019-03-15
Save generated tangents on export

wheeland:
Task-Id: KUE-617


src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -376,8 +376,31 @@ bool vertexSizesForAttributesAreValid(const QVector<Qt3DRender::QAttribute *> &a
...
 namespace MeshParserUtils {
-Qt3DRender::QAttribute *createTangentAttribute(Qt3DRender::QGeometry *geometry, Qt3DRender::QGeometryRenderer::PrimitiveType primitiveType)
+
+Qt3DRender::QAttribute *createNewTangentAttributeIfNeeded(Qt3DRender::QGeometryRenderer *mesh)
 {
+    const auto primitiveType = mesh->primitiveType();
+
+    // only generate tangents for triangles
+    if (primitiveType != Qt3DRender::QGeometryRenderer::Triangles
+            && primitiveType != Qt3DRender::QGeometryRenderer::TriangleStrip
```
lemirep: indent seems wrong, was clang-format run?

src/core/gltf2importer/meshparser_utils.cpp
```diff
@@ -376,8 +376,31 @@ bool vertexSizesForAttributesAreValid(const QVector<Qt3DRender::QAttribute *> &a
 namespace Kuesa {
 namespace GLTF2Import {
 namespace MeshParserUtils {
-Qt3DRender::QAttribute *createTangentAttribute(Qt3DRender::QGeometry *geometry, Qt3DRender::QGeometryRenderer::PrimitiveType primitiveType)
```
lemirep: I would leave createTangentAttribute like it was before and instead create a function hasTangentAttribute. Otherwise we have a function returning nullptr for various conditions (missing geometry, non triangle primitive type ...)

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -352,6 +379,10 @@ Qt3DCore::QEntity *GLTF2Parser::parse(const QByteArray &jsonData, const QString
         return nullptr;
     }
 
+    generateMissingTangents(basePath, rootObject);
```
lemirep: This step should still be controlled by a condition, I can image in some cases we might purposely not want to generate tangents

src/core/gltf2importer/meshparser_utils.cpp
wheeland: right, makes sense!

src/core/gltf2importer/meshparser_utils.cpp
wheeland: fixed, indented it in a way that clang-format is happy with (and how it was before), but I don't actually see the point here. Why would I want to indent it in a way that makes it impossible to distinguish if() condition and the resulting statement, plus not even be able to figure out by looking at the first characters whether this is a || or && ?

src/core/gltf2importer/gltf2parser.cpp
wheeland: we could make this a button for each mesh in the mesh collection?

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -124,6 +136,19 @@ void extractPositionViewDirAndUpVectorFromViewMatrix(const QMatrix4x4 viewMatrix
                          viewMatrix.row(2)[3]);
 }
 
+int addToJsonArray(QJsonObject &object, const QString &name, const QJsonObject &toInsert)
```
jcelerier: there are already some "JSON-util" like functions here : https://github.com/KDAB/kuesa/blob/dev/src/core/gltf2exporter/gltf2utils_p.cpp

maybe they should all be put in some kind of generic json_utils.hpp ?

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -485,6 +519,85 @@ const GLTF2Context *GLTF2Parser::context() const
     return m_context;
 }
 
+void GLTF2Parser::generateMissingTangents(const QString &basePath, QJsonObject &rootObject)
```
jcelerier: I think that this should be refactored to go in the exporter part - maybe by storing a flag in one of the GLTF2Context struct which indicates which mesh had its tangents computed and then by adding a pass here before the draco compression (which might compress the tangent attribute) : https://github.com/KDAB/kuesa/blob/dev/src/core/gltf2exporter/gltf2exporter_p.cpp#L199 ?

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -485,6 +519,85 @@ const GLTF2Context *GLTF2Parser::context() const
...
+                *it = jsonMesh;
+            }
+            rootObject[KEY_MESHES] = jsonMeshes;
+        }
+    }
+
+    // add tangent buffer file
+    if (!tangentBuffer.isEmpty()) {
+        // create file
+        const QLatin1Literal bufferFileName("kuesaTangentBuffers.bin");
```
jcelerier: this function could be useful : https://github.com/KDAB/kuesa/blob/dev/src/core/gltf2exporter/gltf2exporter_p.cpp#L55 (else potentially if loading / saving an asset stuff may be overwritten).

src/core/gltf2importer/gltf2parser.cpp
wheeland: put it into gltf2utils_p.cpp for now

src/core/gltf2importer/gltf2parser.cpp
wheeland: added a button in the MeshWidget for each mess that misses tangents

wheeland:
As of now, I kept this in GLTF2Importer. Doesn't feel like the best place to put it though. Moving everything to GLTF2Exporter doesn't make sense of now, as this one is only created once during export, and we want to see the tangents in action within APE, as materials might depend on them, right?

I guess one way to make it cleaner would be to move the tangent computation itself into GLTF2Context for example, retain some information on what has changed, including a new QByteArray for the newly generated buffer(s), and then in GLTF2Exporter use all that saved info to modify the JSON tree. This would require some slightly awkward duplication of information though, which is not needed as of now.

Trying to figure out what breaks the Draco export now. Does the DracoExportPass expect all data to be stored inside a single buffer file?

mkrus:
could use some unit tests

jalbamon (commented):
Looks OK from a documentation perspective

mkrus (changes requested):
missing unit tests

tests/auto/tangentgenerator/tst_tangentgenerator.cpp
```diff
@@ -0,0 +1,67 @@
...
+        auto meshRenderer = mesh.meshPrimitives[0].primitiveRenderer;
+
+        QVERIFY(Kuesa::GLTF2Import::MeshParserUtils::needsTangentAttribute(meshRenderer));
+
+        // WHEN
+        QVERIFY(Kuesa::GLTF2Import::MeshParserUtils::generateTangentAttribute(meshRenderer, &ctx));
+
+        // THEN
+        QVERIFY(!Kuesa::GLTF2Import::MeshParserUtils::needsTangentAttribute(meshRenderer));
+    }
```
lemirep: Would you mind adding another test case where we load a .gltf file without tangent, generate the tangents, export it as another .gltf and load that exported gltf file and check it contains tangents ?

PR #81, opened 2019-03-15
Move light as child of camera

mkrus:
Will move along with it and not affect viewAll

### DIFF
--- a/tools/assetpipelineeditor/main.qml
+++ b/tools/assetpipelineeditor/main.qml
@@ -41,21 +41,6 @@ Kuesa.SceneEntity {
     readonly property Layer texturePreviewLayer: Layer {}
     readonly property Layer materialPreviewLayer: Layer {}
 
-    // let this point light wander around with the camera to create some shiny lighting
-    Entity {
-        id: pointLightEntity
-        property Camera followedCamera: frameGraph.camera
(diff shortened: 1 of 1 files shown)


wheeland:
Will this not break lighting in case a glTF camera is used?

wheeland:
Yup, just checked. With a glTF camera, the spotlight will not be visible anymore.

tools/assetpipelineeditor/main.qml
```diff
@@ -41,21 +41,6 @@ Kuesa.SceneEntity {
     readonly property Layer texturePreviewLayer: Layer {}
     readonly property Layer materialPreviewLayer: Layer {}
 
-    // let this point light wander around with the camera to create some shiny lighting
-    Entity {
-        id: pointLightEntity
-        property Camera followedCamera: frameGraph.camera
-        parent: followedCamera.parent
```
lemirep: I think the correct fix is actually parent: followedCamera no ?

wheeland: You are right. I forgot that QCamera is just a regular entity with a Transform-component.

wheeland:
Fixed here: https://github.com/KDAB/kuesa/pull/82

PR #82, opened 2019-03-15
Parent PointLight to the camera, not to its parent

wheeland:
This will fix the ViewAll, but keep the lighting intact

### DIFF
--- a/examples/kuesa/car-scene/qml/MainScene.qml
+++ b/examples/kuesa/car-scene/qml/MainScene.qml
@@ -169,14 +169,12 @@ Kuesa.SceneEntity {
     // let this point light wander around with the camera to create some shiny lighting
     Entity {
         id: pointLightEntity
-        property Camera followedCamera: frameGraph.camera
-        parent: followedCamera.parent
+        parent: frameGraph.camera
         components: [
             PointLight {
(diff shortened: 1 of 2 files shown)


mkrus:
why not move it in the file rather than messing with parent?

wheeland:
Because we need the light to be a child of the glTF camera, when it is active. So we need to re-parent it every time the active camera changes.

wheeland:
Update: 
- also use this fix for the car-scene
- make linear and quadratic attenuation 0.0, so that the light has a constant strength, like DirectionalLight
- decrease constantAttenuation to make it brighter

examples/kuesa/car-scene/qml/MainScene.qml
```diff
@@ -169,14 +169,12 @@ Kuesa.SceneEntity {
...
-        property Camera followedCamera: frameGraph.camera
-        parent: followedCamera.parent
+        parent: frameGraph.camera
         components: [
             PointLight {
-                constantAttenuation: 5.0    // don't illuminate too much, but keep the shininess
-            },
-            Transform {
-                translation: pointLightEntity.followedCamera.position
+                constantAttenuation: 1.0
```
mkrus: those are the default I think

PR #86, opened 2019-03-15
Simple gltfViewer

mkrus:
Opens and renders gltf files using just a QWindow, no QML, no widgets.
On macOS this gets associated to .gltf files.
Supports drag & dropping files on the opened window.

Will be useful to debug performance without interference from other rendering.

### DIFF
--- a/tools/assetpipelineeditor/orbitcameracontroller.cpp
+++ b/tools/assetpipelineeditor/orbitcameracontroller.cpp
@@ -294,10 +294,12 @@ void OrbitCameraController::setCamera(Qt3DRender::QCamera *camera)
 
         m_camera = camera;
 
-        if (m_camera)
-            connect(m_camera, &Qt3DCore::QNode::nodeDestroyed, this, [this]() {
-                setCamera(nullptr);
+        if (m_camera) {
+            connect(m_camera, &Qt3DCore::QNode::nodeDestroyed, this, [this, camera]() {
+                if (camera == this->m_camera)
+                    setCamera(nullptr);
             });
+        }
 
         emit cameraChanged();
     }
--- a/tools/gltfViewer/README.md
+++ b/tools/gltfViewer/README.md
@@ -1,0 +1,39 @@
+# Simple glTL Viewer
+
+The gltfViewer application can be used to view glTF files, including
+selecting cameras and playing animations.
+
+Usage: `gltf [-f] [-c name] [-p name] [-l] file.gltf`
+
+## Options
+
+*  `-h, --help`               : Display help.
+*  `-v, --version`            : Display version information.
+*  `-f, --fullScreen`         : Show full screen.
+*  `-p, --play <animations>`  : Play named animation
+*  `-l, --loop`               : Loop animations
+*  `-c, --camera <camera>`    : Use named camera
+
+
+## Operation
+
+The application will render the glTF file passed as an
+argument on the command line.
+
+On macOS, the application is associated to .gltf and .glb files
+so double clicking on one such file in Finder will launch the
+application.
+
+To load another file, simply drag & drop it onto the main window.
+
+The `--fullscreen` argument can be used to render the scene in full screen.
+Using the `Ctrl-F` key combination will toggle the full screen window.
+
+If one or more cameras are present in the file, the first one
+will initially be used to render the scene. A specific camera can
+be selected using the `--camera <name>` argument. Using the `Tab` 
+and `BackTab` keys will cycle through available cameras.
+
+To play an animation, use the `--play <name>` argument. The animation
+will play once. To play it continuously, use the `--loop` argument.
+The `Space` key can be used to pause or resume the current animation.
--- a/tools/gltfViewer/main.cpp
+++ b/tools/gltfViewer/main.cpp
@@ -1,0 +1,229 @@
+/*
+    main.cpp
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2018-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Mike Krus <mike.krus@kdab.com>
+
+    Licensees holding valid proprietary KDAB Kuesa licenses may use this file in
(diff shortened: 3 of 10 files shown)


tools/assetpipelineeditor/orbitcameracontroller.cpp
```diff
@@ -294,10 +294,12 @@ void OrbitCameraController::setCamera(Qt3DRender::QCamera *camera)
 
         m_camera = camera;
 
-        if (m_camera)
-            connect(m_camera, &Qt3DCore::QNode::nodeDestroyed, this, [this]() {
-                setCamera(nullptr);
+        if (m_camera) {
+            connect(m_camera, &Qt3DCore::QNode::nodeDestroyed, this, [this, camera]() {
+                if (camera == this->m_camera)
```
jjcasmar: I dont get why the if

tools/gltfViewer/rotation.json
```diff
@@ -0,0 +1,132 @@
+{
```
jjcasmar: Why this file?

tools/gltfViewer/main.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    ViewerApplication app(argc, argv);
+
+    QCommandLineParser cmdline;
+    cmdline.addHelpOption();
+    cmdline.addVersionOption();
+    cmdline.setApplicationDescription(
+            QObject::tr("\nThis tool renders glTF 2.0 files.\n"
+                        "Example : ./gltfViewer -f my_file.gltf"));
+    cmdline.addPositionalArgument("source", QCoreApplication::translate("main", "gltf 2.0 file to render."));
```
jjcasmar: isnt --input (-i) more natural?

tools/gltfViewer/main.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    QCommandLineOption fullScreenOption({ "f", "fullScreen" }, QObject::tr("show full screen."));
```
jjcasmar: How about adding a key option to go to fullscreen when the app is already open?

tools/gltfViewer/main.qml
```diff
@@ -0,0 +1,97 @@
...
+
+    KuesaViewer.OrbitCameraController {
+        id: controller
+        camera: frameGraph.camera
+        windowSize: _winSize
+    }
+
+    Kuesa.Asset {
+        id: cameraAsset
+        name: root.cameras.names.length ? root.cameras.names[0] : ""
```
jjcasmar: What happens if we have several cameras? Maybe a key to change between cameras would be nice. Also a key to play/stop all animations?

tools/assetpipelineeditor/orbitcameracontroller.cpp
mkrus: hard to see in the PR but this a separate commit with a message explaining.

When the camera is loaded from a gltf file, loading a second file will
destroy the first camera using deleteLater. By that time a new camera
will have been selected and we want to keep using that.

tools/gltfViewer/rotation.json
mkrus: oops, it's part of the kuesa template I used to initially create the project, will remove

tools/gltfViewer/main.cpp
mkrus: -f is for full screen. Guess I should mark it as optional. File is actually a positional argument

tools/gltfViewer/main.cpp
mkrus: will do

tools/gltfViewer/main.qml
mkrus: could use tab to cycle through cameras.
Animations is also possible maybe later. Want to keep this as simple as possible, mostly to use to test on various hardware and with various files.

lemirep:
Does it make sense that it starts up even if no gltf file was specified ? I'd rather have it print the expected usage in that case

mkrus:
Now quits if no source is specified, except on mac since if it's open via the file association (double clicking on a gltf file in Finder) that file is not passed as a command line option but arrives later as a QFileOpenEvent.

mkrus:
> Would probably make sense to add a Readme.md file in the gltfViewer directory explaining how it is expected to work for new users

done

tools/gltfViewer/README.md
```diff
@@ -0,0 +1,39 @@
+# Simple glTL Viewer
+
+The gltfViewer application can be use to previous glTF files, including
```
jalbamon: "preview"  or "view" instead of "previous"?

mkrus: done

tools/gltfViewer/README.md
```diff
@@ -0,0 +1,39 @@
+# Simple glTL Viewer
+
+The gltfViewer application can be use to view glTF files, including
```
jjcasmar: can be used ?

tools/gltfViewer/README.md
```diff
@@ -0,0 +1,39 @@
...
+selecting cameras and playing animations.
+
+Usage: `gltf [-f] [-c name] [-p name] [-l] file.gltf`
+
+## Options
+
+*  `-h, --help`               : Displays help.
+*  `-v, --version`            : Displays version information.
+*  `-f, --fullScreen`         : show full screen.
```
jjcasmar: Shows ?

tools/gltfViewer/README.md
```diff
@@ -0,0 +1,39 @@
...
+*  `-p, --play <animations>`  : play named animation
```
jjcasmar: plays, loops. Or instead of Displays, display. 
Just be sure to use capital case or not for all of them, and third person or not for all of them

tools/gltfViewer/README.md
mkrus: fixed

tools/gltfViewer/README.md
mkrus: done

PR #87, opened 2019-03-16
APE: support picking on meshes in the 3d view

mkrus:


tools/assetpipelineeditor/main.qml
```diff
@@ -81,9 +81,15 @@ Kuesa.SceneEntity {
             }
         },
         ObjectPicker {
+            property point __downPos
+            onPressed: __downPos = pick.position
             onClicked: {
+                if (pick.position !== __downPos)
```
jjcasmar: In C++, __variable are reserved for implementation. Im not used to QML and if is typical to name properties this way. If not, maybe is better to change the name?

mkrus: it's the same idea, it's a convention to indicate the property is reserved for the item's internal implementation and should not be be used from outside

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -282,6 +287,69 @@ void MainWindow::viewAll()
...
+        }
+        if (node) {
+            auto entityIndex = model->index(m_entity->entities(), node->objectName());
+            if (entityIndex.isValid()) {
+                m_ui->collectionBrowser->selectionModel()->select(entityIndex, QItemSelectionModel::ClearAndSelect | QItemSelectionModel::Current);
+                m_ui->collectionBrowser->scrollTo(entityIndex);
+            }
+        }
+    } else {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 13, 0)
```
lemirep: can't componentFromEntity in kuesa_utils_p.h be used here rather than rewriting it?

mkrus: done

mkrus:
Click will pick a mesh, Shift-Click will pick the material, Alt-Click will pick the entity

zakalawe:
> Click will pick a mesh, Shift-Click will pick the material, Alt-Click will pick the entity

Can we have an on-screen prompt about those modes? Tip of the day or close-able help panel or something?

mkrus:
> > Click will pick a mesh, Shift-Click will pick the material, Alt-Click will pick the entity
> 
> Can we have an on-screen prompt about those modes? Tip of the day or close-able help panel or something?

sure, probably as part of a separate change though

zakalawe:
> > > Click will pick a mesh, Shift-Click will pick the material, Alt-Click will pick the entity
> > 
> > 
> > Can we have an on-screen prompt about those modes? Tip of the day or close-able help panel or something?
> 
> sure, probably as part of a separate change though

Oh sure, I just know I never remember these combos without a prompt :)

lemirep:
Could the  key combos be documented somewhere at least?

mkrus:
added some blurb in the top level readme

mkrus:
Also added the ability to customise the background colour

tools/assetpipelineeditor/assetpipelineeditor.pro
```diff
@@ -25,7 +25,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 TEMPLATE = app
-QT += 3dcore 3drender 3dinput 3danimation 3dextras 3dquick widgets kuesa kuesa-private
+QT += 3dcore 3dcore-private 3drender 3drender-private 3dinput 3danimation 3dextras 3dquick widgets kuesa kuesa-private
```
jjcasmar: If I have read the code correctly, the privates are only needed if you dont have 5.13. Can you check that and only add them if that is the case?

mkrus: done

PR #88, opened 2019-03-19
Ensure that the attributes have a name for draco compression

jcelerier:
Currently when exporting generated tangents, no glTF name was set
which would then generate an invalid json while exporting.
This makes this mechanism more robust for future cases.

### DIFF
--- a/src/core/gltf2exporter/dracocompressor_p.cpp
+++ b/src/core/gltf2exporter/dracocompressor_p.cpp
@@ -151,6 +151,35 @@ QByteArray packedDataInBuffer(const QByteArray &inputBuffer,
     return outputBuffer;
 }
 
+QString semanticNameFromAttribute(const Qt3DRender::QAttribute &attribute)
+{
+    auto str = attribute.property("semanticName").toString();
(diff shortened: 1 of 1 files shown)


src/core/gltf2exporter/dracocompressor_p.cpp
```diff
@@ -151,6 +151,35 @@ QByteArray packedDataInBuffer(const QByteArray &inputBuffer,
     return outputBuffer;
 }
 
+QString semanticNameFromAttribute(const Qt3DRender::QAttribute &attribute)
+{
+    auto str = attribute.property("semanticName").toString();
+    if (!str.isEmpty()) {
+        return str;
+    } else {
+        const auto name = attribute.name();
```
lemirep: I guess you could make this more readable by creating a const QHash<QString, QString> and doing a .value lookup on it.

jcelerier: ok, I'll refactor it in another commit since there is pretty similar code elsewhere

PR #89, opened 2019-03-19
Put resources in external files in the common case.

jcelerier:
This looks like it is necessary for arm boards where the linker fails
due to the large dds files.


mkrus:
I can see how it works when running from the build folder, but I think it needs more so that the binary resources get installed properly for the end user

jcelerier:
should be installed in the correct place now

PR #92, opened 2019-03-21
ForwardRenderer: allow rendering into QRenderTarget

wheeland:


src/core/framegraphes/forwardrenderer.cpp
```diff
@@ -589,22 +618,31 @@ void ForwardRenderer::updateTextureSizes()
...
-    } else if (qobject_cast<QOffscreenSurface *>(surface)) {
-        m_resizeConnections.push_back(connect(m_surfaceSelector, &Qt3DRender::QRenderSurfaceSelector::externalRenderTargetSizeChanged, this, &ForwardRenderer::updateTextureSizes));
-    } else {
-        qCWarning(kuesa) << Q_FUNC_INFO << "Unexpected surface type for surface " << surface;
+    if (renderTarget) {
+        Qt3DRender::QAbstractTexture *texture = findRenderTargetTexture(renderTarget, Qt3DRender::QRenderTargetOutput::Color0);
+        m_resizeConnections.push_back(connect(texture, &Qt3DRender::QAbstractTexture::widthChanged, this, &ForwardRenderer::updateTextureSizes));
+        m_resizeConnections.push_back(connect(texture, &Qt3DRender::QAbstractTexture::heightChanged, this, &ForwardRenderer::updateTextureSizes));
+    }
+    else {
```
mkrus: did clang-format allow that? Thought it would be `} else {`

mkrus (commented):
A sample manual test would be great

wheeland:
will be used in the materials demo that I'm preparing to render ground reflections.

mkrus (dismissed):
well it builds and doesn't break the car scene or the forward renderer manual test, so gtg

src/core/framegraphes/forwardrenderer.cpp
```diff
@@ -189,6 +195,18 @@ using namespace Kuesa;
     Holds the size of the external render target.
 */
 
+/*!
```
lemirep: Was this indented with Ctrl+E,R in QtCreator ?

src/core/framegraphes/forwardrenderer.cpp
```diff
@@ -885,6 +935,15 @@ void ForwardRenderer::setRenderSurface(QObject *renderSurface)
     m_surfaceSelector->setSurface(renderSurface);
 }
 
+/*!
+ * Sets the render target that the scene will be rendered into. If set to NULL,
+ * the scene will be rendered into a surface set by setRenderSurface
+ */
+void ForwardRenderer::setRenderTarget(Qt3DRender::QRenderTarget *renderTarget)
+{
+    m_renderTargetSelector->setTarget(renderTarget);
```
lemirep: shouldn't we rebuild the FrameGraph if we had previously set a window ?

lemirep (changes requested):
We need the manual test forward_renderer to have a new tests that allows rendering into a render target, then back to the windows based on a checkbox being checked or something

src/core/framegraphes/forwardrenderer.cpp
wheeland: m_renderTargetSelector::targetChanged is connected to reconfigureFrameGraph() and handleSurfaceChange()

src/core/framegraphes/forwardrenderer.cpp
wheeland: I just copied the formatting of the other property docs

tests/manual/forwardrenderer_scene/OffscreenEntity.qml
```diff
@@ -0,0 +1,74 @@
...
+                    }
+                }
+            }
+        ]
+    }
+
+    TextureMaterial {
+        id: material
+        texture: colorAttachment
+        textureTransform: Qt.matrix4x4
```
lemirep: Qt.matrix4x4(), yields  Unable to assign a function to a property of any type other than var otherwise

lemirep (commented):
Could you also update tst_forwardRenderer to check that setRenderTarget sets and emits properly please?

jalbamon (commented):
Looks OK from a documentation standpoint.

src/core/framegraphes/forwardrenderer.cpp
```diff
@@ -856,14 +900,20 @@ QVector<AbstractRenderStage *> ForwardRenderer::renderStages() const
...
+    Qt3DRender::QRenderTarget *renderTarget = m_renderTargetSelector->target();
 
-    if (auto window = qobject_cast<QWindow *>(surface))
-        size = window->size();
-    else if (qobject_cast<QOffscreenSurface *>(surface))
-        size = m_surfaceSelector->externalRenderTargetSize();
-    else
-        qCWarning(kuesa) << Q_FUNC_INFO << "Unexpected surface type for surface " << surface;
+    if (renderTarget) {
+        Qt3DRender::QAbstractTexture *texture = findRenderTargetTexture(renderTarget, Qt3DRender::QRenderTargetOutput::Color0);
```
jjcasmar: What happens if the render target doesn't have the Color0 attachment?

wheeland: Changed the patch so that in that case it falls back to rendering to a surface (if any is given)

wheeland:
Added an auto test in addition the the manual test. This now tries to fall back to surface rendering and prints a warning.

PR #94, opened 2019-03-21
Fix selection material

mkrus:
Special highlight material still needs.brdfLUT texture. We steel it
from the material of the selected entity. Needs to be done every time
in case another model was loaded and the previous texture was deleted.

Q: what happens when the texture is deleted, does the QParameter know about it?
Does the effect need to track the life time of the texture then? Or should it be 
promoted as a proper parameter like the other maps?

### DIFF
--- a/src/core/metallicroughnesseffect.cpp
+++ b/src/core/metallicroughnesseffect.cpp
@@ -28,6 +28,7 @@
 
 #include "metallicroughnesseffect.h"
 
+#include <Qt3DCore/private/qnode_p.h>
 #include <Qt3DRender/qcullface.h>
 #include <Qt3DRender/qfilterkey.h>
 #include <Qt3DRender/qgraphicsapifilter.h>
@@ -142,6 +143,19 @@ namespace Kuesa {
  */
 
 /*!
+    \property brdfLUT
+
+    brdfLUT references a texture containing lookup tables for the split sum approximation
+    in the PBR rendering. This is used internally by the material.
+
+    When creating an instance of Kuesa::MetallicRoughnessMaterial, users should assign
+    a texture to this property of the effect. A shared instance can be retrieved from
+    the Kuesa::TextureCollection using the name "_kuesa_brdfLUT"
+
+    \since 1.1
+ */
+
+/*!
     \qmltype MetallicRoughnessEffect
     \instantiates Kuesa::MetallicRoughnessEffect
     \inmodule Kuesa
@@ -234,6 +248,19 @@ namespace Kuesa {
     Tone mapping specifies how we perform color conversion from HDR (high
     dynamic range) content to LDR (low dynamic range) content which our monitor
     displays.
+
+    \since 1.1
+ */
+
+/*!
+    \qmlproperty Qt3DRender.AbstractTexture brdfLUT
+
+    brdfLUT references a texture containing lookup tables for the split sum approximation
+    in the PBR rendering. This is used internally by the material.
+
+    When creating an instance of Kuesa::MetallicRoughnessMaterial, users should assign
+    a texture to this property of the effect. A shared instance can be retrieved from
+    the Kuesa::TextureCollection using the name "_kuesa_brdfLUT"
 
     \since 1.1
  */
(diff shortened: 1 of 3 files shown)


src/core/metallicroughnesseffect.cpp
```diff
@@ -718,8 +719,19 @@ void MetallicRoughnessEffect::setToneMappingAlgorithm(MetallicRoughnessEffect::T
...
+    return m_brdfLUTParameter->value().value<Qt3DRender::QAbstractTexture *>();
+}
+
 void MetallicRoughnessEffect::setBrdfLUT(Qt3DRender::QAbstractTexture *brdfLUT)
 {
+    auto d = Qt3DCore::QNodePrivate::get(this);
+    auto current = this->brdfLUT();
+    if (current)
+        d->unregisterDestructionHelper(current);
+    if (brdfLUT)
```
lemirep: you should check if brdfLut has a parent and set it if none was set

mkrus: done

src/core/metallicroughnesseffect.cpp
```diff
@@ -718,8 +719,25 @@ void MetallicRoughnessEffect::setToneMappingAlgorithm(MetallicRoughnessEffect::T
...
+    return m_brdfLUTParameter->value().value<Qt3DRender::QAbstractTexture *>();
+}
+
 void MetallicRoughnessEffect::setBrdfLUT(Qt3DRender::QAbstractTexture *brdfLUT)
 {
+    auto d = Qt3DCore::QNodePrivate::get(this);
+    auto current = this->brdfLUT();
+    if (current) {
+        d->unregisterDestructionHelper(current);
+        if (current->parentNode() == this)
```
lemirep: you don't need to unparent, the old one (who manages its lifetime then ?) You should parent the new brdfLUT to make sure the  Qt3D backend creates it (which we do by setting the parent).

src/core/metallicroughnesseffect.cpp
```diff
@@ -718,8 +719,25 @@ void MetallicRoughnessEffect::setToneMappingAlgorithm(MetallicRoughnessEffect::T
...
+            current->setParent(static_cast<Qt3DCore::QNode *>(nullptr));
+    }
+    if (brdfLUT) {
+        d->registerDestructionHelper(brdfLUT, &MetallicRoughnessEffect::setBrdfLUT, brdfLUT);
+        if (!brdfLUT->parentNode())
+            brdfLUT->setParent(this);
```
lemirep: Usually it's just  if (!brdfLUT->parent()) brdfLUT->setParent(this); In practice I would assume parentNode() == parent() but might be better to align the code with what we do everywhere else.

mkrus: done, although it's kind of odd since the setParent MUST the one that takes a node, the other one being deleted...

src/core/metallicroughnesseffect.cpp
mkrus: ah, yes. But should I delete if the parent is this?

src/core/metallicroughnesseffect.cpp
```diff
@@ -718,8 +719,25 @@ void MetallicRoughnessEffect::setToneMappingAlgorithm(MetallicRoughnessEffect::T
...
+    return m_brdfLUTParameter->value().value<Qt3DRender::QAbstractTexture *>();
+}
+
 void MetallicRoughnessEffect::setBrdfLUT(Qt3DRender::QAbstractTexture *brdfLUT)
 {
+    auto d = Qt3DCore::QNodePrivate::get(this);
+    auto current = this->brdfLUT();
+    if (current) {
+        d->unregisterDestructionHelper(current);
+        if (current->parent() == this)
```
lemirep: Sorry I probably wasn't clear. I was trying to say that the lines 733, 734 shouldn't be needed.

mkrus: what is the effect is the last user of the texture? It's destruction will be delayed until the effect gets deleted. Is that ok?

mkrus: deleted that code

src/core/metallicroughnesseffect.h
```diff
@@ -59,6 +59,7 @@ class KUESASHARED_EXPORT MetallicRoughnessEffect : public Qt3DRender::QEffect
     Q_PROPERTY(bool opaque READ isOpaque WRITE setOpaque NOTIFY opaqueChanged)
     Q_PROPERTY(bool alphaCutoffEnabled READ isAlphaCutoffEnabled WRITE setAlphaCutoffEnabled NOTIFY alphaCutoffEnabledChanged)
     Q_PROPERTY(MetallicRoughnessEffect::ToneMapping toneMappingAlgorithm READ toneMappingAlgorithm WRITE setToneMappingAlgorithm NOTIFY toneMappingAlgorithmChanged REVISION 1)
+    Q_PROPERTY(Qt3DRender::QAbstractTexture *brdfLUT READ brdfLUT WRITE setBrdfLUT NOTIFY brdfLUTChanged REVISION 1)
```
jalbamon: Should add documentation for this property.

mkrus: done

PR #96, opened 2019-03-26
Binary gltf

mkrus:


src/core/gltf2importer/gltf2context.cpp
```diff
@@ -370,6 +370,16 @@ void GLTF2Context::addLocalFile(const QString &file)
     m_localFiles.push_back(file);
 }
 
+QByteArray GLTF2Context::bufferChunk() const
+{
+    return m_bufferChunk;
+}
+
+void GLTF2Context::setBufferChunk(QByteArray &bufferChunk)
```
lemirep: const QByteArray & ? Since you are storing that in a plain QByteArray

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -149,8 +149,38 @@ Qt3DCore::QEntity *GLTF2Parser::parse(const QString &filePath)
...
+    struct GLBHeader {
+        quint32 magic;
+        quint32 version;
+        quint32 length;
+    } glbHeader;
+
+    if (f.read(reinterpret_cast<char *>(&glbHeader), sizeof(glbHeader)) != sizeof(glbHeader)) {
+        isBinary = false;
+    } else {
+        if (glbHeader.magic != 0x46546C67) {
```
lemirep: Would it be cleaner to move the glbHeader check into a dedicated function (that can be more easily tested in a standalone way) ?

tests/auto/gltfparser/tst_gltfparser.cpp
```diff
@@ -503,7 +518,7 @@ private Q_SLOTS:
 
         // THEN
         QVERIFY(res);
-        QCOMPARE(scene.textures()->names().size(), 1);
+        QCOMPARE(scene.textures()->names().size(), 2);
```
lemirep: I don't really understand why this is now 2

lemirep: Ah alright because of the brdfLut

src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -476,6 +506,55 @@ Qt3DCore::QEntity *GLTF2Parser::parse(const QByteArray &jsonData, const QString
...
+            return nullptr;
+        }
+
+        auto data = stream->read(chunkHeader.chunkLength);
+        if (data.size() != chunkHeader.chunkLength) {
+            qCWarning(kuesa()) << "Malformed chunck in glb file";
+            return nullptr;
+        }
+
+        if (chunkHeader.chunkType == 0x4E4F534A) {
```
lemirep: move the hardcoded value to a const variable like we do for JSON keys?

src/core/gltf2importer/gltf2parser.cpp
lemirep: Using a const variable for the magic like we do for the JSON keys would also be nicer

mkrus: done

src/core/gltf2importer/gltf2context.cpp
mkrus: done

src/core/gltf2importer/gltf2parser.cpp
mkrus: done

lemirep:
Tests are adequate. Good from my side

PR #98, opened 2019-03-28
Fix parsing when there is no SceneEntity

jcelerier:


mkrus:
Already covered in https://github.com/KDAB/kuesa/pull/96

PR #101, opened 2019-04-03
Use pregenerated images based on screen dpi

lemirep:
- Add python tool to pregenerate assets for various dpis
- Make StyledSlider/StyledSwitch use pregenerated assets

Task-Id: KUE-598


lemirep:
Simplifies the code a bit, looks alright (not pixel perfect) on most platforms. Should fix the huge number of vertices in QtQuick

mkrus:
Like the other patch, this UI is much too big on iPads
IMG_0015

Are both patches needed?

lemirep:
I'd suggest we get this is even though it's not perfect on all platforms, gets us closers to not having 10000 vertices drawn by QtQuick. I suspect iPad doesn't report the correct dpi and chooses the wrong assets. Can't really fix test for this on my side but I'd assume someone with an ipad could fix that with minimal amount of code changes. Other patch from Nuno is not required.

mkrus:
does it work on Android tablets with hidpi?

lemirep:
DSC01405
Size is adequate on my phone

mkrus:
While I agree it's good in fixing the rendering performance it also breaks scaling, I think because the dpi selection fails to take into account the `Screen.devicePixelRatio`. So on macOS (retina), and iPad, everything is way too large

lemirep:
Ok given we've been stuck on this for almost 2 months, I'll let you take over an fix whatever needs to be for it to work properly on iPad as I can't test it here.

lemirep:
Replaced by Mike's patch

PR #106, opened 2019-04-06
Multiple UV sets

jjcasmar:
glTF 2.0 requires to support at least 2 texture coordinate sets. 

This patch modifies the vertex and fragment shaders to support 2 sets on the glsl side and modify the MetallicRoughnessMaterial to set the correct uniform values to select the correct texture coordinate for each map. 

Note: The UV set selection is done using an if. It could be done also using layers, as the selection is configured per material. Using layers means that we can end up with 125 different combinations. Given that we already have different effects if the material is used for skinning or the primitive using the material have color attribute, we can have 500 different effects. Therefore I decided to use a simple if and avoid the possibility of generating a huge amount of effects. If you prefer, I can switch to layers.


src/core/metallicroughnessmaterial.cpp
```diff
@@ -493,6 +493,11 @@ WrappedSignal<OutputType> wrapParameterSignal(MetallicRoughnessMaterial *self, S
 
 MetallicRoughnessMaterial::MetallicRoughnessMaterial(Qt3DCore::QNode *parent)
     : QMaterial(parent)
+    , m_baseColorUsesTexCoord2(new QParameter(QStringLiteral("baseColorUsesTexCoord2"), bool(false)))
```
lemirep: shouldn't we go for baseColorUsesTexCoord1 instead since on the shaderSize the 2nd set of texture coords is vertexTexCoord1 ?

lemirep:
Propably a good time to instroduce tests for the metallicroughnessmaterial/effect classes that at least check for for proper signal emission on the parameters

src/core/metallicroughnessmaterial.cpp
jjcasmar: you are right, ill change it

jjcasmar:
Added test for the effect

tests/auto/metallicroughnesseffect/tst_metallicroughnesseffect.cpp
```diff
@@ -0,0 +1,222 @@
+/*
+    tst_materialparser.cpp
```
lemirep: tst_metallicroughnesseffectr.cpp

src/core/metallicroughnessmaterial.h
```diff
@@ -53,6 +53,13 @@ namespace Kuesa {
 class KUESASHARED_EXPORT MetallicRoughnessMaterial : public Qt3DRender::QMaterial
 {
     Q_OBJECT
+
+    Q_PROPERTY(bool baseColorUsesTexCoord1 READ isBaseColorUsingTexCoord1 WRITE setBaseColorUsesTexCoord1 NOTIFY baseColorUsesTexCoord1Changed)
```
lemirep: These new properties need to be documented (\property \qmlproperty)

tests/auto/metallicroughnesseffect/tst_metallicroughnesseffect.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2018-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Paul Lemire <paul.lemire@kdab.com>
```
lemirep: Thanks

tests/auto/metallicroughnesseffect/tst_metallicroughnesseffect.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+            effect.setBrdfLUT(texture2D);
+            effect.setBrdfLUT(texture2D);
+
+            QCOMPARE(effect.brdfLUT(), texture2D);
+            QCOMPARE(brdfLUTSpy.count(), 1);
+        }
+    }
+};
+
```
lemirep: This test is fine but it' s not testing for your new properties. You might have to write tst_MetallicRoughnessMaterial I fear

tests/auto/metallicroughnesseffect/tst_metallicroughnesseffect.cpp
jjcasmar: Ive changed it :p

tests/auto/metallicroughnesseffect/tst_metallicroughnesseffect.cpp
jjcasmar: Yes, I was adding that. Added to this patch

lemirep:
Test OK, Documentation OK

PR #109, opened 2019-04-10
car-scene: disable Draco on Android and iOS

lemirep:


jjcasmar:
Should we document that we dont support draco on android/ios? Maybe an option to force it anyway?

lemirep:
Well we do support it on Android. It's just that if you try the car-scene on your phone, the app freezes for 30s while decompressing the assets. So I guess we could document that draco can be slow on lower end devices but not sure we should explicitly say we don't support it on these platforms.

jjcasmar:
Right, bad wording on my side. How about saying that by default is disabled but you can force it with some option?

mkrus:
the obvious way for that (assuming it's even possible) is for the config test to default to `no` on ios/android and the user could force it to on if needed. Although I guest that becomes an all or nothing change rather than just for the demo

lemirep:
I've disabled the autodetection of the draco feature on Android and WinRT as a separate pull request (https://github.com/KDAB/kuesa/pull/116). However I haven' t found a way to make it work for iOS as Qt only knows whether it is configured for Darwin but doesn't allow to filter whether it's iOs or not

mkrus:
think this one can be abandoned in favour of the global fix (which can include iOS)

PR #110, opened 2019-04-10
Adds initial Tone Map example

Kahneolus:
This example has a skybox, point light and helmet asset.
The point light and helmet have toggleable rotations.
The Tone map cannot be set 'live' so a bool just has to be edited
in SceneEntity.qml


examples/kuesa/tonemapping/MainMenu.qml
```diff
@@ -0,0 +1,273 @@
+/*
+    MainMenu.qml
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Mike Krus <mike.krus@kdab.com>
```
jjcasmar: Maybe you want to fix attribution

Kahneolus: Whoops, didn't notice that line in the header. Done now.

jjcasmar:
Why cant you change the tonemap on runtime? It should just trigger a shader effect recompilation

examples/kuesa/tonemapping/ToneMapping.pro
```diff
@@ -0,0 +1,49 @@
+# ToneMapping.pro
```
mkrus: should probably be all lower case like the folder

examples/kuesa/tonemapping/ToneMapping.pro
```diff
@@ -0,0 +1,49 @@
...
+    qml.qrc \
+    assets.qrc
+
+OTHER_FILES += \
+    main.qml \
+    SceneEntity.qml
+
+QML_IMPORT_PATH =
+
+# Default rules for deployment.
```
mkrus: these need fixing, see car-scene.pro

examples/kuesa/tonemapping/assets.qrc
```diff
@@ -0,0 +1,25 @@
+<RCC>
+    <qresource prefix="/">
+        <file>../assets/envmaps/wobbly-bridge/CREDITS.txt</file>
+        <file>../assets/envmaps/wobbly-bridge/wobbly_bridge_4k_cube_irradiance.dds</file>
```
mkrus: these envmap files are missing. Why do you need new ones and not use the ones that were already there?

mkrus (commented):
folder needs to be added to parent project. Ideally controls should be shared in some static lib

zakalawe:
> Why cant you change the tonemap on runtime? It should just trigger a shader effect recompilation

Yep let's definitely do that then!

examples/kuesa/tonemapping/Info-macos.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist SYSTEM "file://localhost/System/Library/DTDs/PropertyList.dtd">
+<plist version="0.1">
+<dict>
+    <key>NSPrincipalClass</key>
+    <string>NSApplication</string>
+    <key>CFBundleIconFile</key>
+    <string>@ICON@</string>
+    <key>CFBundleIdentifier</key>
+    <string>com.kdab.kuesa.assetpipeline</string>
```
mkrus: `com.kdab.kuesa.tonemapping`

examples/kuesa/assets/models/damagedhelmet/rotating_glTF/helmet.qrc
```diff
@@ -0,0 +1,6 @@
+<RCC>
```
mkrus: do we need all the other files if we're only going to use these 2?

examples/kuesa/assets/models/damagedhelmet/rotating_glTF/helmet_images.qrc
```diff
@@ -0,0 +1,8 @@
+<RCC>
```
mkrus: if these are related to the helmet, I would put them all in the same qrc file (I think we split them in the car scene because of the draco version)

examples/kuesa/tonemapping/qml/SceneEntity.qml
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    Kuesa.Asset {
+        id: materialsAsset
+        name: "Material_MR"
+        collection: root3D.materials
+        onNodeChanged: {
+            console.log("material", node)
+            if(node) {
+                if(root3D.tone) {
```
mkrus: you should be able to control this from the UI. Currently it's off by default and the user can't turn it on... what's the point?

mkrus (dismissed):

examples/kuesa/tonemapping/Info-macos.plist
Kahneolus: Done

examples/kuesa/assets/models/damagedhelmet/rotating_glTF/helmet.qrc
Kahneolus: Was just allowing the choice until we had come to a decision. It does seem like the built in animation is the way to go, so I can remove the other versions.

examples/kuesa/tonemapping/qml/SceneEntity.qml
Kahneolus: Done.

examples/kuesa/assets/models/damagedhelmet/rotating_glTF/helmet_images.qrc
Kahneolus: done

mkrus (commented):
All good for me

examples/kuesa/tonemapping/qml/SceneEntity.qml
```diff
@@ -0,0 +1,182 @@
...
+        target: materialsAsset.node
+        property: "toneMappingAlgorithm"
+        value: root3D.metallic
+    }
+
+    Kuesa.GLTF2Importer {
+        id: gltf2importer
+        sceneEntity: root3D
+        assignNames: true
+        source: _assetsPrefix + "RotatingHelmet.gltf"
```
jjcasmar: the scene in this example uses normal mapping but the scene doesnt have tangents. We need to generate them. For that, we need to 
`options.generateTangents: true`
otherwise looks black

jjcasmar (changes requested):

examples/kuesa/tonemapping/qml/SceneEntity.qml
mkrus: done

jjcasmar: on linux I couldnt see the helmet (was black) because it was missing the tangents. could you see it?

mkrus: yes it was fine. Gets dark when you turns on cinematic effect but that's expected I think. Not no difference between with or without tangents

jjcasmar: odd. was failing in linux

jjcasmar (dismissed):

lemirep (approved):
es2 reference should have been _isES2

PR #115, opened 2019-04-23
Add options to the parser

jjcasmar:
Missing tests and doc


src/core/gltf2importer/gltf2importer.cpp
```diff
@@ -208,13 +209,23 @@ using namespace Kuesa;
     \brief if true, assets with no names will be added to collections with default names (default is false)
  */
 
+/*!
+ * \property GLTF2Importer::options
+ * \brief Holds the Kuesa::GLTF2Import::GLTF2Options used to select what to import/generate
```
lemirep: Was this indented in QtC with Ctrl + E, R ?

src/core/gltf2importer/gltf2options.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * \brief Different options to configure the glTF 2.0 importer
+ *
+ * Avalaible options are
+ * \list
+ * \li generateTangents: If true, the importer will generate tangents
+ * for all the primitives which doesn't have tangents.
+ * \endlist
+ */
+
+Kuesa::GLTF2Import::GLTF2Options::GLTF2Options()
```
lemirep: Should the ctor allow to receive a parent in so as to have its lifetime properly managed ?

src/core/gltf2importer/gltf2importer.cpp
jjcasmar: Does nothing. I can indent it by hand so it follows the properties above

src/core/gltf2importer/gltf2options.cpp
jjcasmar: In fact, it feels really wrong the I need to dynamically allocate and manage the lifetime of this object. It should be just a plain structure with the options, but as I need to be able to create it from qml and to emit signals, I need to derive from QObject and make it dynamic. Isnt't there a better approach?

jjcasmar: Refactor so that we dont need to allocate a GLTF2Options object.

src/core/gltf2importer/gltf2options.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <QObject>
+#include <Kuesa/kuesa_global.h>
+
+namespace Kuesa {
+namespace GLTF2Import {
+
+class KUESASHARED_EXPORT GLTF2Options : public QObject
+{
+    Q_OBJECT
```
mkrus: have you tried accessing it from qml? Think you'd need to register properties so you can type: `GLTFImporter { options.generateTangents: true }`

src/core/gltf2importer/gltf2importer.h
```diff
@@ -65,11 +67,14 @@ class KUESASHARED_EXPORT GLTF2Importer : public Qt3DCore::QNode
...
     Kuesa::SceneEntity *sceneEntity() const;
     bool assignNames() const;
+    Kuesa::GLTF2Import::GLTF2Options *options();
+    const Kuesa::GLTF2Import::GLTF2Options *options() const;
 
 public Q_SLOTS:
     void setSource(const QUrl &source);
     void setSceneEntity(Kuesa::SceneEntity *sceneEntity);
     void setAssignNames(bool assignNames);
+    void setOptions(const Kuesa::GLTF2Import::GLTF2Options &options);
```
mkrus: having a setter for a `CONSTANT` property seems weird. Why do you need it?
Isn't writing `importer->options->set...()` nicer?

mkrus (changes requested):

src/core/gltf2importer/gltf2importer.h
jjcasmar: Was a decision to make it a bit more cppish. You can access directly the itnernal object as you suggest or use the setter, which just copies the value internally

mkrus: hm, superfluous I think, but ok. But the ownership rules are a bit weird...

If I call `setOptions(options)` and `options` has no parent, I would expect that the importer takes ownership of the options object. It of course doesn't because the data is copied. But it's still different from elsewhere in Qt3d/Kuesa

src/core/gltf2importer/gltf2options.h
jjcasmar: right

src/core/gltf2importer/gltf2importer.h
jjcasmar: Well, you are passing a const ref, no way to take ownership, as typical in c++. I can remove it if you prefer it.

mkrus: ah right, oops, read it as a pointer. I just find the API weird

src/quick/imports/kuesaplugin.cpp
```diff
@@ -82,6 +82,7 @@ void KuesaPlugin::registerTypes(const char *uri)
     qmlRegisterType<Kuesa::Skybox>(uri, 1, 0, "Skybox");
     qmlRegisterType<Kuesa::Asset>(uri, 1, 0, "Asset");
     qmlRegisterExtendedType<Kuesa::AnimationPlayer, Kuesa::AnimationPlayerItem>(uri, 1, 0, "AnimationPlayer");
+    qmlRegisterType<Kuesa::GLTF2Import::GLTF2Options>(uri, 1, 1, "GLTF2Options");
```
lemirep: should be qmlRegisterUncreatableType, we don't want people to be able to create GLTF2Options nodes in QML

lemirep (dismissed):

src/core/gltf2importer/gltf2importer.h
jjcasmar: I think we can leave it as it is. We have the cpp and qml idioms.

tools/assetpipelineeditor/mainwindow.cpp
```diff
@@ -492,11 +499,13 @@ void MainWindow::openSettings()
 {
     QSettings settings;
     auto clearColor = settings.value("clearColor", QColor(Qt::gray)).value<QColor>();
+    auto generateTangents = settings.value("generateTangents", false).value<bool>();
 
     SettingsDialog dlg(this);
     dlg.setSelectionColor(m_assetInspector->meshInspector()->selectionColor());
     dlg.setClearColor(clearColor);
     dlg.setDefaultClearColor(settings.value("defaultClearColor", true).toBool());
+    dlg.setGenerateTangents(generateTangents);
```
mkrus: pretty sure you can use `m_generateTangent` here

(clearColour is a special case because you don't always want to use the user provided one)

jjcasmar: done

mkrus (approved):

PR #116, opened 2019-04-24
Disable auto detection of Draco on Android and WinRT

lemirep:
Qt doesn't offer a way to distinguish between desktop Darwin platforms
and mobile ones (iOS). We therefore leave it enabled for Darwin.

### DIFF
--- a/src/core/configure.json
+++ b/src/core/configure.json
@@ -21,6 +21,7 @@
     "features": {
         "draco": {
             "label": "Draco",
+            "autoDetect": "!config.uikit && !config.android && !config.winrt",
             "output": [
                 "publicFeature",
                 "feature",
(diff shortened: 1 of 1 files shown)


jjcasmar:
Looks fine to me

src/core/configure.json
```diff
@@ -21,6 +21,7 @@
     "features": {
         "draco": {
             "label": "Draco",
+            "autoDetect": "!config.android && !config.winrt",
```
mkrus: You can use `!config.uikit` (which excludes iOS, tvOS, watchOS but keeps macOS)

mkrus:
Added the disabling of Draco in Apple mobile platforms

lemirep:
Looks good to me but github doesn't let me approve it

jcelerier:
can't either

mkrus (approved):

PR #117, opened 2019-04-25
gles linter

TimoBuskeKdab:


lemirep (commented):
Some files are missing copyright headers, they need to be set so that we know if that's KDAB code or not

TimoBuskeKdab:
ah, right. thanks! added them.

lemirep (approved):

PR #119, opened 2019-04-27
Introduce MetallicRoughnessTechnique

mpersano:
Factor out some of the repeated code in MetallicRoughnessEffect.

### DIFF
--- a/src/core/metallicroughnesseffect.cpp
+++ b/src/core/metallicroughnesseffect.cpp
@@ -47,6 +47,106 @@ using namespace Qt3DRender;
 
 namespace Kuesa {
 
+class MetallicRoughnessTechnique : public Qt3DRender::QTechnique
+{
+public:
+    explicit MetallicRoughnessTechnique(Qt3DCore::QNode *parent = nullptr);
+
+    void setZFillShaderProgram(Qt3DRender::QShaderProgram *shaderProgram);
+    void setVertexShader(const QUrl &url);
+
+    QStringList enabledLayers() const;
+    void setEnabledLayers(const QStringList &layers);
+    void setOpaque(bool opaque);
+    void setCullingMode(QCullFace::CullingMode mode);
+
+private:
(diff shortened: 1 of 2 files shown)


src/core/metallicroughnesstechnique.cpp
```diff
@@ -0,0 +1,361 @@
...
+    , m_opaque(true)
+    , m_alphaCutoffEnabled(false)
+    , m_toneMappingAlgorithm(MetallicRoughnessEffect::Reinhard)
+    , m_invokeInitVertexShaderRequested(false)
+    , m_backFaceCulling(new QCullFace(this))
+    , m_metalRoughShaderBuilder(new QShaderProgramBuilder(this))
+    , m_metalRoughShader(new QShaderProgram(this))
+{
+    const auto enabledLayers = QStringList{ QStringLiteral("noBaseColorMap"),
+                                                QStringLiteral("noMetalRoughMap"),
```
lemirep: Is indent correct here ? usually clang-format doesn't behave like that

src/core/metallicroughnesstechnique_p.h
```diff
@@ -0,0 +1,130 @@
+/*
+    metallicroughnesstechnique_p.h
+
+    This file is part of Kuesa.
+
+    Copyright (C) 2018-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+    Author: Paul Lemire <paul.lemire@kdab.com>
```
lemirep: thx :)

jjcasmar:
I disagree in the way this patch refactor the technique. Before the patch, we only had one copy of each property, now we have 3 copies of each property, one for each technique. 
I also think its confusing and error prone to set the property for each technique, but only emit a signal from the GL3 technique. 
Isnt there a way to refactor the common bits of the technique, but leave properties on the effect?

mpersano:
@jjcasmar Good points, updated the patch.

PR #124, opened 2019-05-08
Ui fixes (non scaling)

mkrus:
This "fixes" scaling by not scaling. Fits desktop (retina or not), tablet, and phone (kind of, needs scrolling). This fixes performance issues by using images.

Needs to be tested on something like the tegra. If it's too big, I don't think we should scale but use specific UI.

Started moving components to separate lib for reused. Needs more work to move the entire panel (without the menu content)


lemirep:
Looks good on my phone

PR #130, opened 2019-05-15
Warn missing tangent

lemirep:


src/core/gltf2importer/gltf2parser.cpp
```diff
@@ -913,50 +914,70 @@ void GLTF2Parser::generateTreeNodeContent()
...
+                            if (metalRoughMat != nullptr && metalRoughMat->normalMap() != nullptr) {
+                                Q_ASSERT(renderer->geometry());
+                                const QVector<Qt3DRender::QAttribute *> attributes = renderer->geometry()->attributes();
+                                bool hasTangentAttribute = std::find_if(attributes.cbegin(),
+                                                                        attributes.cend(),
+                                                                        [](Qt3DRender::QAttribute *attr) {
+                                                                            return attr->name() == Qt3DRender::QAttribute::defaultTangentAttributeName();
+                                                                        }) != attributes.cend();
+                                if (!hasTangentAttribute)
+                                    qWarning() << QStringLiteral("Primitive %1 is trying to use Material %2 which does normal mapping even though it defines no tangent attribute. This can result in incorrect rendering, please consider generating tangents.")
```
mkrus: this is going to be weird if the material or primitive doesn't have a name. Ideally could use index but I don't think we have access to that info easily at this point...

mkrus (approved):

src/core/gltf2importer/gltf2parser.cpp
lemirep: I could point out the primitive and/or material are nameless if that can be better

PR #132, opened 2019-05-15
Unhide createAttribute from DRACO ifdef

jjcasmar:
createAttribute doesn't use any draco stuff and is used in the non draco code path. Therefore, it fails to link in Android, where draco is disabled.


jcelerier:
looks good to me

src/core/gltf2importer/meshparser.cpp
```diff
@@ -501,8 +501,8 @@ MeshParser::geometryMorphTargetsFromJSON(Qt3DRender::QGeometry *geometry,
         for (const MorphTargetAttribute &morphTargetAttribute : morphTarget.attributes) {
             const QString semanticName = morphTargetAttribute.name;
             const QString attributeName = QStringLiteral("%1_%2")
-                                          .arg(standardAttributeNameFromSemantic(semanticName))
-                                          .arg(i + 1);
+                                                  .arg(standardAttributeNameFromSemantic(semanticName))
```
mkrus: clang-format I presume. Weird

PR #133, opened 2019-05-22
Fix shaders android

jjcasmar:
Several fixes for es2/3. 
Tested in my android phone and several scenes work fine, including the ghost one that was reported to fail in android but not on desktop. 

Requires a patch in Qt3D which is already merged in 5.12 but not yet in dev branch
https://codereview.qt-project.org/c/qt/qt3d/+/261631


jjcasmar:
WhatsApp Image 2019-05-22 at 13 23 27

PR #134, opened 2019-05-23
Use brdfLut texture in texture collection for pbr effect

jjcasmar:
When the material component is added to an entity, it searches for a SceneEntiy and if found, uses the brdfLut texture in the texture collection.

### DIFF
--- a/src/core/metallicroughnessmaterial.cpp
+++ b/src/core/metallicroughnessmaterial.cpp
@@ -27,6 +27,7 @@
 */
 
 #include "metallicroughnessmaterial.h"
+#include "kuesa_p.h"
 
 #include <Qt3DRender/qtexture.h>
 #include <Qt3DRender/qparameter.h>
@@ -35,6 +36,8 @@
 
 #include "metallicroughnesseffect.h"
 #include "morphcontroller.h"
+
+#include "sceneentity.h"
 
 QT_BEGIN_NAMESPACE
 
@@ -122,6 +125,11 @@ WrappedSignal<OutputType> wrapParameterSignal(MetallicRoughnessMaterial *self, S
     nodes of the shader graph and will trigger a recompilation of the shader.
     It is recommended to create the variations of the material that you need
     once up front at application initialisation time.
+
+    \note The effect behind this material needs a brdfLut texture. If the brdfLut
+    texture in the effect is not set and the material is added to a subtree of a
+    Kuesa::SceneEntity, it will look for the Kuesa::SceneEntity and will use the
+    brdfLut texture in the texture collection.
 */
 
 /*!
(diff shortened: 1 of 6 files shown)


src/core/sceneentity.h
```diff
@@ -135,6 +135,8 @@ class KUESASHARED_EXPORT SceneEntity : public Qt3DCore::QEntity
     Qt3DRender::QTextureLoader *m_brdfLUT;
 };
 
+SceneEntity *findParentSceneEntity(Qt3DCore::QEntity *entity);
```
mkrus: shouldn't this be in `kuesa_utils_p.h`?
And if it stays here, shouldn't it be a static public member of SceneEdity?
Or if it stays a separate function, shouldn't it have `KUESASHARED_EXPORT`?

src/core/metallicroughnessmaterial.cpp
```diff
@@ -816,6 +830,22 @@ void MetallicRoughnessMaterial::setToneMappingAlgorithm(MetallicRoughnessEffect:
...
+
+    // Looks on the entity tree until we find a SceneEntity
+    // while we find the SceneEntity, get the brdfLut texture and add it to the effect
+    auto *sceneEntity = Kuesa::findParentSceneEntity(entity);
+    if (!sceneEntity) {
+        qCWarning(kuesa) << "A material has been added to a subtree without a SceneEntity";
+        return;
+    }
+
+    m_effect->setBrdfLUT(sceneEntity->texture(QLatin1String("_kuesa_brdfLUT")));
```
mkrus: since we're have access to the SceneEntity here, could we have the texture as one of it's properties and avoid having it in the collection?

mkrus (changes requested):

src/core/metallicroughnessmaterial.cpp
jjcasmar: Modified to use the textuer as a property, instead of adding it to the texture collection

src/core/sceneentity.h
jjcasmar: Changed. Now is a static method of SceneEntity.

src/core/sceneentity.cpp
```diff
@@ -361,4 +357,22 @@ Qt3DCore::QNode *SceneEntity::transformForEntity(const QString &name)
     return componentFromEntity<Qt3DCore::QTransform>(e);
 }
 
+Qt3DRender::QAbstractTexture *SceneEntity::brdfLut() const
```
mkrus: this now needs documentation, sorry :)

jjcasmar: that was unexpected :)
added

